Issue #113, opened 2015-06-04
Some unit tests are failed


vbauer:
I've tried to build project using `mvn clean install` and got the following errors:

```
shouldConvertToDateFromString(org.jooby.internal.ParamConverterTest)  Time elapsed: 0.02 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at org.jooby.internal.ParamConverterTest.shouldConvertToDateFromString(ParamConverterTest.java:122)
```

```
Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.385 sec - in org.jooby.internal.PostConstructTest
Running org.jooby.internal.PreDestroyTest
ERROR [2015-06-05 04:18:34,930] [main] org.jooby.Managed.stop() resulted in error
java.io.IOException: null
	at org.easymock.internal.MockInvocationHandler.invoke(MockInvocationHandler.java:46) ~[na:3.2]
	at org.easymock.internal.ObjectMethodsFilter.invoke(ObjectMethodsFilter.java:94) ~[na:3.2]
	at com.sun.proxy.$Proxy26.stop(Unknown Source) ~[na:na]
	at org.jooby.internal.LifecycleProcessor.onPreDestroy(LifecycleProcessor.java:98) ~[na:na]
	at org.jooby.internal.PreDestroyTest.lambda$shouldNOTFailOnException$304(PreDestroyTest.java:174) [na:na]
	at org.jooby.internal.PreDestroyTest$$Lambda$229/1556635644.run(Unknown Source) [na:na]
	at org.jooby.MockUnit.run(MockUnit.java:136) [na:na]
	at org.jooby.internal.PreDestroyTest.shouldNOTFailOnException(PreDestroyTest.java:173) [na:na]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_25]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_25]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_25]
	at java.lang.reflect.Method.invoke(Method.java:483) ~[na:1.8.0_25]
[...]

PR #114, opened 2015-06-06
#113: Fix unit test "shouldConvertToDateFromString"

vbauer:


vbauer:
Btw, I'm not sure that throwing exceptions during execution of correct tests is a good idea (https://travis-ci.org/jooby-project/jooby/builds/65704281). Maybe it is possible to prevent them?

PR #240, opened 2015-12-11
Switch case for patterns

franklinexpress:
Would it be cleaner to add more patterns in the future using a switch?


jknack:
Hi,

 Thanks for the pull request. What other pattern do you have in mind?

PR #284, opened 2016-02-24
Correct typo in  'Routes' documentation

authchir:
`Route.Err` does not seems to exist. It seems that `Err` was meant.


jknack:
appreciate. Release 1.0 is comming but still need to review all the doc.. so please keep sending these fixes and/or feel free to improve/change the doc when doesn't make sense or if something is missing.

Issue #273, opened 2016-02-16
Conform locale matching with RFC4647


authchir:
I just got strange results from the `Result.locale()` function and took a quick look at the implementation. If my understanding is correct, it seems that the [current locale detection algorithm][1] is a little naïve:

1. Set the default locale from the configuration file as the request's locale.
2. If the header `Accept-Language` is present, split naïvely the value to create a new locale.

It would be interesting if the locale matching was conforming to the [RFC4647][2], wich is already implemented in Java's [`Locale`][3].

[1]:https://github.com/jooby-project/jooby/blob/3d2e2994ff09979fddbec44ea91b90e623928a8e/jooby/src/main/java/org/jooby/internal/RequestImpl.java#L101-L102
[2]:http://tools.ietf.org/html/rfc4647
[3]:https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#LocaleMatching

jknack:
I built a multi-language app (traditional chinese, japanese, spanish and portugal) without major issues.

But yea is probably not enough. Are you able to send a pull request?

Thanks

authchir:
I am just starting with Jooby, but I can give it a try. It would be an opportunity to get to know the framework better.

jknack:
got it, last time I check that I found some diff between ```-``` used by the browser, or ```_``` used by Locale.

will review and improve it, don't worry :)

can you add an example with the error or problem you get?

thanks.

authchir:
```Java
import java.util.Locale;
import org.jooby.Jooby;

public class Main extends Jooby {
  {
    get("/", req -> {
      Locale l = req.locale();
      return String.format(
        "Language: %s\nCountry: %s\nVariant: %s\nScript: s\n",
        l.getLanguage(),
        l.getCountry(),
        l.getVariant(),
        l.getScript());
    });
  }

  public static void main(final String[] args) throws Exception {
    new Main().start(args);
  }
}
```

```
$ curl localhost:8080 -H 'Accept-Language:"de-DE,de;q=0.8,fr-CA;q=0.7,fr;q=0.5,en-CA;q=0.3,en;q=0.2"'
Language: "de
Country: DE,DE;Q=0.8,FR
Variant: CA;q=0.7,fr;q=0.5,en
Script: s
```

authchir:
[...]

PR #298, opened 2016-03-13
Refactor the Request locale API

authchir:
Following the addition of:
 - `List<Locale> locales(BiFunction<List<Locale.LanguageRange>, List<Locale>, List<Locale>> filter`
 - `Locale locale(BiFunction<List<Locale.LanguageRange>, List<Locale>, Locale> filter)`

`Locale locale(Iterable<Locale> locales)` and `Locale locale(Locale... locales)`
were made redundant by the new two functions. This use case can be implemented
in the following way:

    List<Locale> supportedLocales = ...
    Locale locale = req.locale((ranges, locales) ->
      Locale.lookup(ranges, supportedLocales));

`List<Locale> locales()` is just an shorthand for `locales(Locale::filter)`.

`Locale locale()` simply returns the first element of `locales` or the default
locale as per `app.lang` if there are no matches. This function is expected to
be the most used one.

Issue #273

### DIFF
--- a/coverage-report/src/test/java/org/jooby/issues/Issue273.java
+++ b/coverage-report/src/test/java/org/jooby/issues/Issue273.java
@@ -1,5 +1,6 @@
 package org.jooby.issues;
 
+import java.util.Arrays;
 import java.util.Locale;
 
 import org.jooby.test.ServerFeature;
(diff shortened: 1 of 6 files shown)


jooby/src/main/java/org/jooby/internal/LocaleUtils.java
```diff
@@ -50,10 +51,14 @@ public static Locale parseOne(final String value) {
...
-    return range.stream()
-        .sorted(Comparator.comparing(Locale.LanguageRange::getWeight).reversed())
-        .collect(Collectors.toList());
+    try {
+      List<Locale.LanguageRange> range = Locale.LanguageRange.parse(value);
+      return range.stream()
+          .sorted(Comparator.comparing(Locale.LanguageRange::getWeight).reversed())
+          .collect(Collectors.toList());
+    } catch (IllegalArgumentException e) {
+      return Collections.emptyList();
```
jknack: why this? users will never know there was a problem

authchir: This was to allow absent or erroneous `Accept-Language` (see three first cases of `Issue273.noMatch()`). Do you think that only the absent case should be handle?

jknack: think so yea, absence = empty list and wrong value as exception

authchir:
Ok, I removed the `try-catch`. The absence of the `Accept-Language` was already handle in `RequestImpl.locale()` and `RequestImpl.locales()`.

PR #311, opened 2016-03-22
Jooby QueryDSL-JDBC support

DerDackel:
I kinda hope this isn't unsolicited as I've seen, you guys seem to be working on a QueryDSL module of your own, using QueryDSL with JPA. I've recently implemented a small service based on Jooby and tried out QueryDSL on plain JDBC specifically because I had relatively few interactions with the DB and wanted to avoid the bload of JPA.

...and suddenly I ended up with everything to exract a generic Jooby module for integrating QueryDSL over JDBC.
I've borrowed heavily from the jOOQ module to get the layout in shape. Biggest hurdle was the actual lack of any SQL Dialect detection, but turns out, nobody seems to do anything smarter than examining the DB URL string.

Maybe you've already got this on the backburner or decide it is redundant to have, but I'd like to hear your input on this!

### DIFF
--- a/jooby-querydsl/pom.xml
+++ b/jooby-querydsl/pom.xml
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>jooby-project</artifactId>
+        <groupId>org.jooby</groupId>
+        <version>0.17.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>jooby-querydsl</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.querydsl</groupId>
+            <artifactId>querydsl-core</artifactId>
+            <version>${querydsl.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.querydsl</groupId>
+            <artifactId>querydsl-sql</artifactId>
+            <version>${querydsl.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jooby</groupId>
+            <artifactId>jooby-jdbc</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>org.jooby</groupId>
+            <artifactId>jooby</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <classifier>tests</classifier>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.easymock</groupId>
+            <artifactId>easymock</artifactId>
+            <scope>test</scope>
+        </dependency>
+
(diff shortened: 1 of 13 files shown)


jknack:
Hey thanks for this. I never used querydsl with raw jdbc, but yea why not, right?

I will make some comments on your pull and I do have this already:

1. can you amend your commits and submit just one?
2. you're right about dialect detection and what you did is OK..but if you look at the jdbc module there is a dialect detection (similar to yours) but much more extensive and tested. In Jdbc.java, there is a ```dbtype``` variable that holds the db type
3. is there a way of using querydsl without code generation?

The querydsl for JPA it is just the setup for code generation: http://jooby.org/doc/querydsl

Again, thanks for this pull.

jooby-querydsl/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>jooby-querydsl</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.querydsl</groupId>
+            <artifactId>querydsl-core</artifactId>
+            <version>4.0.9</version>
```
jknack: please create a ```querydsl.version``` variable in parent ```pom.xml``` and use it here

jooby-querydsl/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.querydsl</groupId>
+            <artifactId>querydsl-sql</artifactId>
+            <version>4.0.9</version>
```
jknack: same here: ```${querydsl.version}```

jooby-querydsl/src/main/java/org/jooby/querydsl/QueryDSL.java
```diff
@@ -0,0 +1,149 @@
...
+
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * <h1>QueryDSL</h1>
+ * <p>
+ * SQL abstraction provided by <a href="http://www.querydsl.com"QueryDSL</a> using plain JDBC underneath.
```
jknack: bad link you need to close the ```<a````

jooby-querydsl/src/main/java/org/jooby/querydsl/QueryDSL.java
```diff
@@ -0,0 +1,149 @@
...
+ *   <a href="http://www.querydsl.com/static/querydsl/latest/reference/html_single/#d0e725">QueryDSL documentation.</a>
+ * </p>
+ *
+ *
+ * @since 0.17.0-SNAPSHOT
+ * @author sjackel
+ */
+public class QueryDSL extends Jdbc {
+
+  private BiConsumer<Configuration, Config> callback;
```
jknack: unused var. you never call it

jooby-querydsl/src/main/java/org/jooby/querydsl/SQLDialectDetector.java
```diff
@@ -0,0 +1,69 @@
...
+package org.jooby.querydsl;
+
+import com.querydsl.sql.*;
+
+import javax.sql.DataSource;
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.SQLException;
+
+public interface SQLDialectDetector {
```
jknack: think can be replaced it with ```Jdbc.dbtype``` is private now, you need to make it accesible

DerDackel:
Thanks for the feedback! Will fix on the morrow!

DerDackel:
I almost forgot: There's some DBs that are supported QueryDSL that aren't yet listed in this module, as well as the advanced functionality of building and setting your own dialect templates, but I thought I'd get an initial, usable module done first.

Issue #356, opened 2016-04-11
`AssetHandler` overridden in in `Request.assets(String, AssetHandler)`


authchir:
I was trying to configure some aspects of my assets and encountered a strange situation where some configuration (i.e. `maxAge`) was working bug others (i.e. `etag` and `lastModified`). Here is a small example tuned for testing:

```Java
assets("/assets/css/**", new AssetHandler("/").etag(false).lastModified(false).maxAge(Duration.ofDays(1).getSeconds()));
assets("/assets/img/**", new AssetHandler("/").etag(false).lastModified(true).maxAge(Duration.ofDays(14).getSeconds()));
assets("/assets/js/**",  new AssetHandler("/").etag(true).lastModified(false).maxAge(Duration.ofDays(2).getSeconds()));
assets("/i18n/**",       new AssetHandler("/").etag(true).lastModified(true).maxAge(Duration.ofDays(7).getSeconds()));
```

The value given for `etag` and `lastModified` is always ignored. I searched in the implementation of `AssetHandler` but everything seems fine. I then inspected the implementations of the different `Request.assets` functions and found that the values of `AssetHandler.cdn`, `AssetHandler.lasModified` and `AssetHandler.etag` are always overridden with the value given in the configuration file:

```Java
@Override
public Route.Definition assets(final String path, final AssetHandler handler) {
  AssetProxy router = new AssetProxy();
  Route.Definition asset = new Route.Definition("GET", path, router);
  on("*", conf -> {
    router.fwd(handler
      .cdn(conf.getString("assets.cdn"))
      .lastModified(conf.getBoolean("assets.lastModified"))
      .etag(conf.getBoolean("assets.etag")));
  });
  return appendDefinition(asset);
}
```

I guess it was so written, so that the default `AssetHandler` created by the other functions would be initialized, but it actually prevents us to specify a fine-grained configuration for different assets. I think this behaviour should be moved to the other `assets` functions.

What are your thoughts?

PS: Quite of topic but, since the framework is already Java 8, I would suggest using the nice [`java.time.Duration`][1] class for the `maxAge` attribute.

[1]: https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html

jknack:
Agreed, will review, cleanup and fix... forgot to do it before 1.0.0.CR1 ... think I can remove AssetProxy now.

```Duration``` makes perfect sense will add that too. Good suggestion

authchir:
[...]

PR #362, opened 2016-04-20
Refactor `AssetHandler` according to issue #356

authchir:


jooby/src/main/java/org/jooby/handlers/AssetHandler.java
```diff
@@ -185,11 +187,20 @@ public AssetHandler cdn(final String cdn) {
   }
 
   /**
-   * @param maxAge Set the cache header max-age value in seconds.
+   * @param maxAge Set the cache header max-age value.
    * @return This handler.
    */
-  public AssetHandler maxAge(final long maxAge) {
-    this.maxAge = maxAge;
+  public AssetHandler maxAge(final Optional<Duration> maxAgeOpt) {
```
jknack: don't think we need this method: ```Optional<Duration>```. If we allow a user to set this... why allow an possible empty object? We shouldn't.

Also and to keep backward compatibility.. we should keep the ```long``` version as it was.

jooby/src/main/java/org/jooby/handlers/AssetHandler.java
```diff
@@ -253,9 +264,9 @@ private void doHandle(final Request req, final Response rsp, final Asset asset)
     }
 
     // cache max-age
-    if (maxAge > 0) {
-      rsp.header("Cache-Control", "max-age=" + maxAge);
-    }
+    maxAgeOpt.ifPresent(d -> {
```
jknack: we might have a bug here, default value for maxAge is ```-1```. You change maxAge to be an optional now and think it will be always present, which cause to ALWAYS set the Cache-Control header to a max-age=1

can you confirm? and fix it?

jknack:
Looks good! just a few minor comments

jooby/src/main/java/org/jooby/handlers/AssetHandler.java
authchir: > If we allow a user to set this... why allow an possible empty object? We shouldn't.

The initial value being to be empty, setting a duration activate the feature. The purpose of this method is to allow the user to either activate or deactivate the feature.

> Also and to keep backward compatibility.. we should keep the long version as it was.

Ok, no problem with that.

jooby/src/main/java/org/jooby/handlers/AssetHandler.java
authchir: > You change maxAge to be an optional now and think it will be always present, which cause to ALWAYS set the Cache-Control header to a max-age=1

The definition of the field is as followed:

```Java
private Optional<Duration> maxAgeOpt = Optional.empty();
```

So there is initially no value, which means that, by default, the field is not set.

jooby/src/main/java/org/jooby/handlers/AssetHandler.java
jknack: Yea, that is my point: it is disabled by default. So think we should allow to just enable, calling this method with an empty optional isn't useful.

jooby/src/main/java/org/jooby/handlers/AssetHandler.java
jknack: There is an ```assets.cache.maxAge = -1``` in ```jooby.conf```...think Assets.java use this and there we are probably setting a cache-control header with max-age= -1... not sure

Anyway, I'm close to release 1.0.0.CR3 and want to get all these in. Let me know when is done so I can merge this.

Thanks

jknack:
will merge and finish

PR #400, opened 2016-06-11
Used the PebbleEngine#cacheActive method to enable/disable Pebble tem…

kshep92:
Using `PebbleEngine#templateCache(null)` by itself doesn't disable template caching, the `PebbleEngine#cacheActive()` method needs to be called. See [here](https://github.com/PebbleTemplates/pebble/blob/master/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java#L418-L428)

### DIFF
--- a/jooby-pebble/src/main/java/org/jooby/pebble/Pebble.java
+++ b/jooby-pebble/src/main/java/org/jooby/pebble/Pebble.java
@@ -186,6 +186,7 @@ public class Pebble implements Jooby.Module {
   public void configure(final Env env, final Config conf, final Binder binder) {
     /** Template cache. */
     String mode = env.name();
+	pebble.cacheActive(!mode.equals("dev"));
     if (mode.equals("dev") || conf.getString("pebble.cache").isEmpty()) {
       pebble.templateCache(null);
     } else {


jooby-pebble/src/main/java/org/jooby/pebble/Pebble.java
```diff
@@ -186,6 +186,7 @@ public Pebble doWith(final Consumer<PebbleEngine.Builder> callback) {
   public void configure(final Env env, final Config conf, final Binder binder) {
     /** Template cache. */
     String mode = env.name();
+	pebble.cacheActive(!mode.equals("dev"));
```
jknack: move this line inside the existing checks:

```java
if (mode.equals("dev") || conf.getString("pebble.cache").isEmpty()) {
  pebble.templateCache(null);
  pebble.cacheActive(false);
} else {
  pebble.templateCache(CacheBuilder.from(conf.getString("pebble.cache")).build());
  pebble.cacheActive(true);
}
```

jknack:
Awesome! But think you need to update: PebbleTest.java

kshep92:
Here's a kicker, using this method disables caches globally for both templates and tags. We need to be smart about where we insert this logic otherwise users might find that they set a `pebble.tagCache` value, but caches are still disabled because the caches were turned off in the preceding check. 

Here are my two suggestions for dealing with this:

1. A separate method for dealing with caches
2. A new config item that is read to determine if to use caches or not. Something like `pebble.cacheEnabled`

I'm partial toward option 2.

jknack:
as we agreed on gitter.. will merge and finalize

PR #459, opened 2016-08-25
Add Locale Wrapper to AssetCompilerTest

DerDackel:
I had another Whoopsie with a test for number formatting on German locale. I wrote a little wrapper function that'll set the default locale for a block of test code and will reset the original default locale afterwards.
Split the test itself between a US and a GERMAN locale section.

### DIFF
--- a/jooby-assets/src/test/java/org/jooby/assets/AssetCompilerTest.java
+++ b/jooby-assets/src/test/java/org/jooby/assets/AssetCompilerTest.java
@@ -7,10 +7,7 @@ import static org.junit.Assert.assertTrue;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.nio.file.Paths;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import org.jooby.Asset;
 import org.jooby.MediaType;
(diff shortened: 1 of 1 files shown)


jknack:
Thanks! locales are always a pain! will made some small comments

jooby-assets/src/test/java/org/jooby/assets/AssetCompilerTest.java
```diff
@@ -203,4 +206,16 @@ private Config conf(final String path, final String env) {
...
+
+  private void withLocale(Locale locale, Block block) {
+    Locale systemDefault = Locale.getDefault();
+    Locale.setDefault(locale);
+    block.run();
+    Locale.setDefault(systemDefault);
+  }
+
+  @FunctionalInterface
+  private interface Block {
```
jknack: `Runnable` just fine, no?

jooby-assets/src/test/java/org/jooby/assets/AssetCompilerTest.java
```diff
@@ -203,4 +206,16 @@ private Config conf(final String path, final String env) {
         .withValue("assets.env", ConfigValueFactory.fromAnyRef(env))
         .withValue("assets.charset", ConfigValueFactory.fromAnyRef("UTF-8"));
   }
+
+  private void withLocale(Locale locale, Block block) {
+    Locale systemDefault = Locale.getDefault();
+    Locale.setDefault(locale);
+    block.run();
```
jknack: let's be safe and run this inside a `try/finally`

DerDackel:
*facepalm* Yeah, could've thought of those two myself! Should not submit PRs after quitting time :laughing:

jknack:
Thanks!

PR #462, opened 2016-08-26
Add named globbing using "**:name" or "{name:**}"

DerDackel:
Enhances the routing path matcher by allowing the binding of globs as a named parameter using the following syntax:

```java
get("/foo/**:tail", req -> {
  doSomething(req.params("tail")); // GET/foo/bar/baz -> tail == "bar/baz"
});

get("/foo/{tail:**}", req -> {
  doSomething(eq.params("tail")); // same
});

get("/foo/**:inbetween/end", req -> {
  doSomething(eq.params("inbetween")); // GET/foo/bar/baz/end -> inbetween == "bar/baz"
});

get("/foo/{inbetween:**}/end", req -> {
  doSomething(eq.params("inbetween")); // same
});
```

### DIFF
--- a/jooby/src/test/java/org/jooby/internal/RoutePatternTest.java
+++ b/jooby/src/test/java/org/jooby/internal/RoutePatternTest.java
@@ -139,6 +139,104 @@ public class RoutePatternTest {
   }
 
   @Test
+  public void anyNamed() {
+    new RoutePathAssert("GET", "com/**:rest")
+        .matches("GET/com/test.jsp", vars -> assertEquals("test.jsp", vars.get("rest")))
+        .matches("GET/com/a/test.jsp", vars -> assertEquals("a/test.jsp", vars.get("rest")))
+        .matches("GET/com/", vars -> assertEquals("", vars.get("rest")))
+        .butNot("GET/com")
+        .butNot("GET/test.jsp");
+  }
+
+  @Test
+  public void anyNamedSpring() {
+    new RoutePathAssert("GET", "com/{rest:**}")
+        .matches("GET/com/test.jsp", vars -> assertEquals("test.jsp", vars.get("rest")))
+        .matches("GET/com/a/test.jsp", vars -> assertEquals("a/test.jsp", vars.get("rest")))
+        .matches("GET/com/", vars -> assertEquals("", vars.get("rest")))
+        .butNot("GET/com")
+        .butNot("GET/test.jsp");
+  }
+
+  @Test
+  public void anyNamedInner() {
+    new RoutePathAssert("GET", "com/**:rest/bar")
+        .matches("GET/com/foo/bar", (vars) -> assertEquals("foo", vars.get("rest")))
+        .matches("GET/com/a/foo/bar", (vars) -> assertEquals("a/foo", vars.get("rest")))
+        .butNot("GET/com/foo/baz")
+        .butNot("GET/test.jsp");
+  }
+
+  @Test
+  public void anyNamedInnerSpring() {
+    new RoutePathAssert("GET", "com/{rest:**}/bar")
+        .matches("GET/com/foo/bar", (vars) -> assertEquals("foo", vars.get("rest")))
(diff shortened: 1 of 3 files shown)


jknack:
damn! you are fast!!

jooby/src/test/java/org/jooby/internal/RoutePatternTest.java
```diff
@@ -139,6 +139,72 @@ public void any2() {
...
+        .matches("GET/com/a/foo/bar/moo/baz", (vars) -> {
+          assertEquals("a/foo", vars.get("first"));
+          assertEquals("moo/baz", vars.get("second"));
+        })
+        .butNot("GET/com/foo/baz")
+        .butNot("GET/test.jsp");
+  }
+
+  @Test
+  public void anyNamedMultiSpring() {
```
rodrigotechera: Is it worth testing any name multi with mixed syntax? Something like `new RoutePathAssert("GET", "com/**:first/bar/{second:**}")`

jooby/src/test/java/org/jooby/internal/RoutePatternTest.java
```diff
@@ -139,6 +139,72 @@ public void any2() {
   }
 
   @Test
+  public void anyNamed() {
+    new RoutePathAssert("GET", "com/**:rest")
+        .matches("GET/com/test.jsp", vars -> assertEquals("test.jsp", vars.get("rest")))
+        .matches("GET/com/a/test.jsp", vars -> assertEquals("a/test.jsp", vars.get("rest")))
+        .matches("GET/com/", vars -> assertEquals("", vars.get("rest")))
+        .butNot("GET/test.jsp");
```
rodrigotechera: Is it worth testing `matches("GET/com", vars -> assertEquals("", vars.get("rest")))`? 
Notice that it doesn't has the last `/` in one of your test, it seems to me another corner case but not 100% sure.

DerDackel: I added that case deliberately, wanted to make sure the regex wasn't accidently including leading slashes on an empty match. Can change the behavior, but I find it to be more consistent, as that "/" is basically the leading slash of an empty path.

jooby/src/test/java/org/jooby/internal/RoutePatternTest.java
DerDackel: Can't hurt! Added!

jooby/src/test/java/org/jooby/internal/RoutePatternTest.java
```diff
@@ -139,6 +139,72 @@ public void any2() {
...
+  }
+
+  @Test
+  public void anyNamedInner() {
+    new RoutePathAssert("GET", "com/**:rest/bar")
+        .matches("GET/com/foo/bar", (vars) -> assertEquals("foo", vars.get("rest")))
+        .matches("GET/com/a/foo/bar", (vars) -> assertEquals("a/foo", vars.get("rest")))
+        .butNot("GET/com/foo/baz")
+        .butNot("GET/test.jsp");
```
rodrigotechera: Is it worth testing `matches("GET/com/foo/bars/bar", (vars) -> assertEquals("foo/bars", vars.get("rest")))`? Or `matches("GET/com/foo/abar/bar", (vars) -> assertEquals("foo/bars", vars.get("rest")))`
Again, it seems to me another corner case but not 100% sure.

jooby/src/test/java/org/jooby/internal/RoutePatternTest.java
rodrigotechera: Hey, not saying you should remove any of your test. They seems to be good, just suggesting/asking for more test if they worth adding them.

DerDackel: I was just explaining the reasoning for that test case, while misreading a char in your example myself :) Will take care of that one too!

PR #463, opened 2016-08-27
add Caffeine cache support

ben-manes:
This is derived from the Guava module. The changes were,
* Replace Guava types with Caffeine's (`Cache`, `LoadingCache`, `CaffeineSpec`)
* Add AsyncLoadingCache support
* Update documentation, modules, etc.

Note that `AsyncLoadingCache` does not extend the Cache interface due to some minor differences. It would ideally be a `LoadingCache<K, CompletableFuture<V>>`, but that would require allowing a null return value (as Caffeine does not throw an exception on a cache miss). To support this type, `doWithAsync` was added along with `AsyncCallback`. The private `callback` instance has weaker type safety, but is internal so this has no negative effect. That approach allowed for making the most minimal changes from the original Guava code.

### DIFF
--- a/jooby-caffeine/src/main/java/org/jooby/caffeine/CaffeineCache.java
+++ b/jooby-caffeine/src/main/java/org/jooby/caffeine/CaffeineCache.java
@@ -1,0 +1,319 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.jooby.caffeine;
+
+import static java.util.Objects.requireNonNull;
+
+import java.lang.reflect.ParameterizedType;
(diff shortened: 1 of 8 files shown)


jooby-caffeine/README.md
```diff
@@ -0,0 +1,128 @@
+# caffeine
```
jknack: move this file to `md/doc/caffeine/caffeine.md`. The release tool will create this file later.

jooby-caffeine/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+* ```Cache```
+
+## dependency
+
+```xml
+<dependency>
+ <groupId>org.jooby</groupId>
+ <artifactId>jooby-caffeine-cache</artifactId>
+ <version>1.0.0.Final</version>
```
jknack: replace `1.0.0.Final` with `{{version}}`

jooby-caffeine/src/main/java/org/jooby/caffeine/CaffeineCache.java
```diff
@@ -0,0 +1,318 @@
...
+ *
+ * <pre>
+ * {
+ *   get("/", req {@literal} {
+ *     Cache cache = req.require("session", Cache.class);
+ *   });
+ * }
+ * </pre>
+ *
+ * @author edgar
```
jknack: you're the author ;)

jooby-caffeine/src/main/java/org/jooby/caffeine/CaffeineSessionStore.java
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * <pre>
+ * {
+ *   get("/", req {@literal} {
+ *     Cache cache = req.require("session", Cache.class);
+ *   });
+ * }
+ * </pre>
+ *
+ * @author edgar
```
jknack: same here

jknack:
looks great! just did some minor comments

ben-manes:
Done. Added myself as secondary author, since your code was changed in only minor ways when adapting. I also just released a new minor version with an enhancement in a sibling module. That might cause a build break until Sonatype staging fully replicates to Central.

PR #500, opened 2016-10-13
Cleaned up http addresses.

nikos:
Fixing double occurrences.


jknack:
thanks @nikos and good timing bc I'm about to review and improve doc. So,if you found something odd, missing examples, or any other way you can help me to improve doc will be great.

Issue #522, opened 2016-10-25
[jooby-jdbc] Use / configure com.impossibl.postgres.jdbc.PGDriver failed


ludmiloff:
I'm trying to connect to postgres database via com.impossibl.postgres.jdbc.PGDriver.
Unfortunately, the drive has no url property to set up. When I set db.url = .... in application.conf, jooby-jdbc failed with following:

Caused by: java.lang.RuntimeException: Property url does not exist on target class com.impossibl.postgres.jdbc.PGDataSource
	at com.zaxxer.hikari.util.PropertyElf.setProperty(PropertyElf.java:131)
	at com.zaxxer.hikari.util.PropertyElf.lambda$setTargetFromProperties$0(PropertyElf.java:57)
	at java.util.Hashtable.forEach(Hashtable.java:878)
	at com.zaxxer.hikari.util.PropertyElf.setTargetFromProperties(PropertyElf.java:52)
	at com.zaxxer.hikari.pool.PoolBase.initializeDataSource(PoolBase.java:315)
	at com.zaxxer.hikari.pool.PoolBase.<init>(PoolBase.java:108)
	at com.zaxxer.hikari.pool.HikariPool.<init>(HikariPool.java:99)
	at com.zaxxer.hikari.HikariDataSource.<init>(HikariDataSource.java:71)
	at org.jooby.jdbc.Jdbc.configure(Jdbc.java:363)
	at org.jooby.jdbc.Jdbc.configure(Jdbc.java:334)
	at org.jooby.Jooby.install(Jooby.java:4709)

When I omit the db.url property, jooby-jdbc fails again, this time with:
Caused by: com.typesafe.config.ConfigException$Missing: No configuration setting found for key 'db.url'
	at com.typesafe.config.impl.SimpleConfig.findKeyOrNull(SimpleConfig.java:152)
	at com.typesafe.config.impl.SimpleConfig.findOrNull(SimpleConfig.java:170)
	at com.typesafe.config.impl.SimpleConfig.findOrNull(SimpleConfig.java:176)
	at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:184)
	at com.typesafe.config.impl.SimpleConfig.find(SimpleConfig.java:189)
	at com.typesafe.config.impl.SimpleConfig.getString(SimpleConfig.java:246)
	at org.jooby.jdbc.Jdbc.configure(Jdbc.java:351)
[...]

PR #527, opened 2016-10-27
jooby-jdbc configure callback fix

ludmiloff:
Execute callback just before HikariDataSource creation, passing HikariConfig and Config objects


jknack:
Fix #522

PR #529, opened 2016-11-01
Upgrade Elasticsearch client to 5.0.0

spinscale:
* Remove capability to start a node. This should be a client implementation
* Remove classes needed to exposed the client
* Expose RestClient as client

The RESTClient is pretty raw, but does not depend on Elasticsearch core anymore.
Also it is version independent as it uses HTTP.

Mocking of the whole implementation does not work, because the RestClientBuilder is
marked as final.


jknack:
The new module looks great, but we lost the ability of embedded an elasticsearch instance in Jooby.

Don't think that is a good idea.... I know that super powered elasticsearch application won't use this model... but still think is useful for small/medium size app, no?

spinscale:
That was intentional. Elasticsearch is not supposed to run embedded with more recent versions, starting in 2.x. Also it has the security manager enabled by default in 2.x - which for sure will not work with Jooby with all the different dependencies. Last but not least there is another feature called JarHell, which checks that classes do not exist twice in the class path - I am sure this would lead to troubles.

I'd consider Elasticsearch more as a system to store data like cassandra, mongo or mysql - and thus this implementation should be a client only - even though this requires some more work for smaller sites as it is required to set up a dedicated Elasticsearch instance.

jknack:
Got it. Thanks for explaining me. What if we keep the old elasticsearch module and we just add/create a new one: `elasticsearch5`?

For example, I made the mistake of replace `hbm` module with Hibernate 5 (previous was Hibernate 4)... and then I created a `hbm4` module. I don't like what I did and prefer to be release a new module with version as part of the name (elasticsearch release go to fast)

To be clear:

1. leave the elasticsearch alone (probably rename to `elasticsearch-embedded`
2. create a new module: `elasticsearch5`

What do you think ?

spinscale:
I'm happy to create a new module. However supporting other folks software that has been released more than two and a half years ago sounds pretty tedious to me. As Elasticsearch 1.x and 2.x is not BWC compatible on the wire protocol (which is used by the node client), you are limiting your users to the 1.x Elasticsearch version - which should not be used anywhere, as it is not maintained anymore.

The good news with this release is, you dont have to care too much anymore for fast Elasticsearch releases, because of using the HTTP client in this PR (and yeah, I'm happy to upgrade it with the next minor release).

It's totally your call of course (and I am happy to help where needed), but I personally would decide against having two modules, just to not have the maintenance hassle.

jknack:
Give it another thought .... and agreed with you (this time).

Issue #535, opened 2016-11-01
Swagger/jooby-maven-plugin:spec: NPE when using Constructor in App


jschneider:
If there is a constructor in the App, I get that exception:

org.jooby.Err: Server Error(500)
	at org.jooby.internal.HttpHandlerImpl.handleErr(HttpHandlerImpl.java:391)
	at org.jooby.internal.HttpHandlerImpl.cleanup(HttpHandlerImpl.java:371)
	at org.jooby.internal.HttpHandlerImpl.handle(HttpHandlerImpl.java:312)
	at org.jooby.internal.netty.NettyHandler.channelRead0(NettyHandler.java:92)
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:372)
	at io.netty.channel.AbstractChannelHandlerContext.access$600(AbstractChannelHandlerContext.java:38)
	at io.netty.channel.AbstractChannelHandlerContext$7.run(AbstractChannelHandlerContext.java:363)
	at io.netty.util.concurrent.DefaultEventExecutor.run(DefaultEventExecutor.java:66)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:873)
	at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:144)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.IllegalStateException: Error while processing class com.mycompany.App
	at org.jooby.spec.RouteProcessor.processInternal(RouteProcessor.java:298)
	at org.jooby.spec.RouteProcessor.process(RouteProcessor.java:133)
	at org.jooby.internal.swagger.SwaggerBuilder.build(SwaggerBuilder.java:111)
	at org.jooby.swagger.SwaggerUI.lambda$install$2(SwaggerUI.java:255)
	at org.jooby.Route$OneArgHandler.handle(Route.java:1727)
	at org.jooby.internal.RouteImpl.handle(RouteImpl.java:108)
	at org.jooby.internal.RouteChain.next(RouteChain.java:70)
	at org.jooby.Route$Chain.next(Route.java:2074)
	at org.jooby.internal.HttpHandlerImpl.handle(HttpHandlerImpl.java:304)
	... 9 common frames omitted
[...]

PR #537, opened 2016-11-02
Feature/#535 npe with swagger

jschneider:
Solves the NPE as explained in #535

### DIFF
--- a/jooby-spec/src/main/java/org/jooby/spec/RouteProcessor.java
+++ b/jooby-spec/src/main/java/org/jooby/spec/RouteProcessor.java
@@ -220,6 +220,9 @@ public class RouteProcessor {
        * Find out app node.
        */
       Node appNode = new AppCollector().accept(unit, ctx);
+      if (appNode == null) {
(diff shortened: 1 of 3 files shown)


jooby-spec/src/main/java/org/jooby/spec/RouteProcessor.java
```diff
@@ -220,6 +220,9 @@
        * Find out app node.
        */
       Node appNode = new AppCollector().accept(unit, ctx);
+      if (appNode == null) {
+        throw new IllegalStateException("No app node found");
```
jknack: What if we raise a `Empty constructor not found` (or similar)

jknack:
Thanks. Can you amend your commit and send just 1 commit?

Thanks again

jooby-spec/src/main/java/org/jooby/spec/RouteProcessor.java
jschneider: Yes. Maybe. I have no idea what that null check does ;-)

Issue #538, opened 2016-11-02
Swagger: Does not accept static inner classes


jschneider:
I will create a pull request with a unit test reproducing the problem

jschneider:
https://github.com/jooby-project/jooby/pull/539

PR #539, opened 2016-11-02
add unit test to reproduce #538

jschneider:
Reproduces issue #538


jknack:
Fixed. Thanks for the test

PR #540, opened 2016-11-02
Feature/#536 swagger ignore response param

jschneider:


jknack:
Good, but here is my feedback:

1. we should change the spec: https://github.com/jooby-project/jooby/tree/master/jooby-spec if we change there clients of spec (swagger and raml) will just work

2. Add `org.jooby.Session`, `org.jooby.Route` but also think we should skip parameters annotated with `org.jooby.mvc.Local` and `org.jooby.mvc.Flash`

Let me know if you have questions

jschneider:
Don't exactly understand what changes you expect in the spec. 

Session and Route are skipped now.
I have improved the tests to check for @Local and @Flash.
But I have no idea how these annotations could be detected...

jknack:
closing this pull request, change need to be done in jooby-spec not in swagger/raml.

I will take care of this.

Thanks

PR #607, opened 2017-01-17
Add support for external health check registry.

krisztiankocsis:
Rationale: Some libraries (like HikariCP) do not provide a healtcheck instance but instead require access to a healthcheck registry to operate. While we can use a different healthcheck registry and delegate the checks, it whould be a good idea to see them in the output of "/sys/healthchecks".


krisztiankocsis:
There is an error in a unit test, I will fix it within a few days.

jknack:
Sorry, I missed your comment and merged it :(

I will fix the test.

Thanks

PR #608, opened 2017-01-17
Server worker executor

krisztiankocsis:
Rational: When doing async tasks with RxJava or other frameworks, it would be a good idea to reuse the server's worker threads.

### DIFF
--- a/jooby/src/main/java/org/jooby/Jooby.java
+++ b/jooby/src/main/java/org/jooby/Jooby.java
@@ -108,6 +108,7 @@ import org.jooby.internal.LocaleUtils;
 import org.jooby.internal.ParameterNameProvider;
 import org.jooby.internal.RequestScope;
 import org.jooby.internal.RouteMetadata;
+import org.jooby.internal.ServerExecutorProvider;
 import org.jooby.internal.ServerLookup;
 import org.jooby.internal.ServerSessionManager;
 import org.jooby.internal.SessionManager;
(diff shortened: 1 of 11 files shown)


jooby/src/main/java/org/jooby/Jooby.java
```diff
@@ -2365,6 +2366,24 @@ public Jooby executor(final Executor executor) {
    * Default executor runs each task in the thread that invokes {@link Executor#execute execute},
    * that's a Jooby worker thread. A worker thread in Jooby can block.
    *
+   * @param name Name of the executor.
+   * @param provider Provider for the executor.
+   * @return This jooby instance.
+   */
+  public Jooby executor(final String name, final Class<? extends javax.inject.Provider<Executor>> provider) {
```
jknack: how useful is a this method? I mean does it need to be public?

jooby/src/main/java/org/jooby/internal/ServerExecutorProvider.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @Override
+  public Executor get() {
+    return executor;
+  }
+
+  static class ServerHolder {
+
+    @Inject(optional = true) Server server = null;
```
jknack: don't think it can be null, right?

jknack:
Good work!

jooby/src/main/java/org/jooby/Jooby.java
krisztiankocsis: Not necessarily, I will make it private

jooby/src/main/java/org/jooby/internal/ServerExecutorProvider.java
krisztiankocsis: I have to make **Server** optional with this workaround because in **Issue576** unit test, there is no instance bound to **Server** class. In real environments it won't be null.

jknack: Fair enough, will later at this.

krisztiankocsis:
Done, should I create a new PR or it is possible the update this one?

jknack:
Update this one. Thanks.

krisztiankocsis:
It seems that this is not my fault, my checks run correctly:

Failed tests: 
  ShouldLoadSessionFromCookieIDFeature.shouldLoadSessionFromCookieID:30->lambda$shouldLoadSessionFromCookieID$2:33 expected:<[1]> but was:<[2]>

jknack:
@krisztiankocsis right, that tests fails randomly, ignore it.

PR #609, opened 2017-01-17
Executors pool size config

krisztiankocsis:
Rational: It would be a good idea to have an alternative configuration style for custom executors like:

executors {
  custom {
    type = fixed
    size = ${runtime.processors-x2}
  }
}

### DIFF
--- a/jooby-executor/src/main/java/org/jooby/exec/Exec.java
+++ b/jooby-executor/src/main/java/org/jooby/exec/Exec.java
@@ -27,6 +27,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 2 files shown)


jooby-executor/src/main/java/org/jooby/exec/Exec.java
```diff
@@ -360,23 +362,38 @@ private static ForkJoinWorkerThreadFactory fjwtf(final String name) {
...
+      options.put(type, config.containsKey("size") ?
+        Integer.parseInt(config.get("size").toString()) : n);
+      options.put("daemon", config.containsKey("daemon") ?
+        Boolean.parseBoolean(config.get("daemon").toString()) : daemon);
+      options.put("asyncMode", config.containsKey("asyncMode") ?
+        Boolean.parseBoolean(config.get("asyncMode").toString()) : false);
+      options.put("priority", config.containsKey("priority") ?
+        Integer.parseInt(config.get("priority").toString()) : priority);
+    } else {
+      System.out.println("Bere");
```
jknack: remove println statement

krisztiankocsis:
Working on it.

jooby-executor/src/main/java/org/jooby/exec/Exec.java
krisztiankocsis: Done

krisztiankocsis:
Done

PR #615, opened 2017-01-23
Fix loading of environment-dependent Logback config file on classpath.

krisztiankocsis:


jooby/src/main/java/org/jooby/Jooby.java
```diff
@@ -3196,11 +3197,17 @@ static String logback(final Config conf) {
     if (conf.hasPath("logback.configurationFile")) {
       logback = conf.getString("logback.configurationFile");
     } else {
+      String env = conf.hasPath("application.env") ? conf.getString("application.env") : null;
+      if (env != null) {
+        URL classpathLogback = Thread.currentThread().getContextClassLoader().getResource("logback." + env + ".xml");
```
jknack: Why `.getContextClassLoader`? We usually use: `getClass().getResource("/logback." + env + ".xml")`

jknack:
Two more comments:

1. We usually prefer file over classpath for logback, bc file let's us turn on/off logging at runtime. So the classpath check should be done after the file system check, not before as you did.

2. We need a unit test

jooby/src/main/java/org/jooby/Jooby.java
krisztiankocsis: If I remember correctly, I read some articles that stated that it is a better option in servlet containers and the same in standalone applications. Since Jooby supports running in a servlet container I thought that this should be used.

krisztiankocsis:
1. That's right, I will change it
2. Working on it

jooby/src/main/java/org/jooby/Jooby.java
jknack: Read the opposite, that we should avoid context class loader. Also, existing code use the `getClass().getResource` method and think we should keep using it.

jknack:
@krisztiankocsis Two more related comments:

1. Do you plan to send more pull requests? I'm close to release `1.0.2`. Let me know if I should wait for more pull requests.

2. Want to be a team member? I need someone to review my code too :)

jooby/src/main/java/org/jooby/Jooby.java
krisztiankocsis: Ok, 1 minute and ready

krisztiankocsis:
1. Probably no in the next two week
2. Since we already have a running app using Jooby and planning to use it in the upcoming projects, I'm interested in becoming a member if I can help in someway

I'm not so familiar with mocking so I need some help on how to mock that URL class required to test this PR.

jknack:
Cool. I will ping you to review my code :)

Leave the test to me, I will merge later and add the test. 

One more comment about pull request/code convention. I prefer to `amed` commits while doing code reviews. This way the commit history is clear and clean.

krisztiankocsis:
Ok

krisztiankocsis:
Hi!

Jooby.class.getResource() does not work for some reason, contextual class loader works.
I cannot you getClass(), it is a static method.

Have you any idea why doesn't it work?

jknack:
Use an absolute path with: `Jooby.class.getResource()` like `/logback[.env].xml`? (note the leading `/`)

krisztiankocsis:
Updated, it works

PR #625, opened 2017-02-02
Fixed #621 - Assets don't work on HTTPS with Netty

### LINKED ISSUE #621: Assets don't work on HTTPS with Netty
When using Netty, file assets don't work when using HTTPS.

The problem is in the NettyResponse, on this line:
      ctx.write(new DefaultFileRegion(channel, offset, count));
The DefaultFileRegion will be sent to the pipeline and first passes through the HttpObjectEncoder. This one will not do any conversion on a FileRegion.
Next the message is passed on to the SslHandler. This one checks if the message is a ByteBuf and otherwise simply returns failure.

Thus, only the header is sent out for assets over HTTPS.

nielsavonds:


jooby-netty/src/main/java/org/jooby/internal/netty/NettyResponse.java
```diff
@@ -184,11 +189,28 @@ public void send(final FileChannel channel, final long offset, final long count)
     rsp.headers().set(headers);
     ChannelHandlerContext ctx = this.ctx;
     ctx.channel().attr(NettyRequest.NEED_FLUSH).set(false);
+
+    // add chunker
+    ChannelPipeline pipeline = ctx.pipeline();
+    if (pipeline.get("chunker") == null) {
```
jknack: We need `chunker` only if we are doing SSL, right? We should probably test that and add when necessary.

jooby-netty/src/main/java/org/jooby/internal/netty/NettyResponse.java
```diff
@@ -184,11 +189,28 @@ public void send(final FileChannel channel, final long offset, final long count)
...
+      pipeline.addAfter("codec", "chunker", new ChunkedWriteHandler());
+    }
+
+    // Create the cunked input here already, to properly handle the IOException
+    final HttpChunkedInput chunkedInput = new HttpChunkedInput(new ChunkedNioFile(channel, offset, count, 8192));
```
jknack: Similar question here, you always create a chunked input but we only need while doing ssl.

jooby-netty/src/main/java/org/jooby/internal/netty/NettyResponse.java
nielsavonds: That's true, though it won't hurt to have it there anyway. I even thought about moving it away here and adding it to NettyPipeline, but that's more of a design choice. I'll leave that to you :) Feel free, to only add it when there's SSL involved.

jooby-netty/src/main/java/org/jooby/internal/netty/NettyResponse.java
nielsavonds: This is the same instance, no? I don't think I added it twice...

nielsavonds: True, the main reason I put this outside the event loop was to be able to properly handle the IOException.

jknack: Got it. Thanks.

PR #628, opened 2017-02-07
File watcher module

jknack:
# file watcher

Watches for file system changes or event. It uses a watch service to monitor a directory for changes so that it can update its display of the list of files when files are created or deleted.

## dependency

```xml
<dependency>
 <groupId>org.jooby</groupId>
 <artifactId>jooby-file watcher</artifactId>
 <version>{{version}}</version>
</dependency>
```

## usage

Watch ```mydir``` and listen for create, modify or delete event on all files under it:

```java
{
  use(new FileWatcher()
    .register(Paths.get("mydir"), (kind, path) -> {
      log.info("found {} on {}", kind, path);
    });
  );
}
```

## file handler

You can specify a {@link FileEventHandler} instance or class while registering a path:

Instance:

```java
{
  use(new FileWatcher()
    .register(Paths.get("mydir"), (kind, path) -> {
      log.info("found {} on {}", kind, path);
    });
  );
}
```

Class reference:

```java
{
  use(new FileWatcher()
    .register(Paths.get("mydir"), MyFileEventHandler.class)
  );
}
```

Worth to mention that ```MyFileEventHandler``` will be provided by Guice.

## options

You can specify a couple of options at registration time:

```java
{
  use(new FileWatcher(
    .register(Paths.get("mydir"), MyFileEventHandler.class, options -> {
      options.kind(StandardWatchEventKinds.ENTRY_MODIFY)
             .recursive(false)
             .includes("*.java");
    });
  ));
}
```

1. Here we listen for {@link StandardWatchEventKinds#ENTRY_MODIFY} (we don't care about create or delete).

2. We turn off recursive watching, only direct files are detected.

3. We want ```.java``` files and we ignore any other files.

## configuration file

In addition to do it programmatically, you can do it via configuration properties:

```java
{
  use(new FileWatcher()
    .register("mydirproperty", MyEventHandler.class)
  );
}
```

The ```mydirproperty``` property must be present in your ```.conf``` file.

But of course, you can entirely register paths from ```.conf``` file:

```
filewatcher {
  register {
    path: "mydir"
    handler: "org.example.MyFileEventHandler"
    kind: "ENTRY_MODIFY"
    includes: "*.java"
    recursive: false
  }
}
```

Multiple paths are supported using array notation:

```java
filewatcher {
  register: [{
    path: "mydir1"
    handler: "org.example.MyFileEventHandler"
  }, {
    path: "mydir2"
    handler: "org.example.MyFileEventHandler"
  }]
}
```

Now use the module:

```java
{
  use(new FileWatcher());
}
```

The [FileWatcher]({{defdocs}}/filewatcher/FileWatcher.html) module read the ```filewatcher.register``` property and setup everything.

### DIFF
--- a/jooby-filewatcher/src/main/java/org/jooby/filewatcher/FileEventOptions.java
+++ b/jooby-filewatcher/src/main/java/org/jooby/filewatcher/FileEventOptions.java
@@ -1,0 +1,177 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.jooby.filewatcher;
+
+import com.sun.nio.file.SensitivityWatchEventModifier;
+
+import java.io.IOException;
+import java.nio.file.*;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchEvent.Modifier;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+
+import static java.nio.file.StandardWatchEventKinds.*;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Allow to customize a file watch handler. You can listen for particular event kinds, apply a glob
+ * filter, recursive listening for changes and set watcher priority or modifier.
+ * <p>
+ * Default filter watch recursively listen for all the available kinds using a <code>HIGH</code>
+ * modifier.
+ * <p>
+ * This class can't be instantiated by client code. Intances of this class are provided via
+ * configuration callbacks. See
+ * {@link FileWatcher#register(Path, Class, java.util.function.Consumer)}
+ *
+ * @author edgar
+ * @since 1.1.0
+ */
+public class FileEventOptions {
+
+  private List<WatchEvent.Kind<Path>> kinds = new ArrayList<>();
+
+  static final PathMatcher TRUE = new PathMatcher() {
+
+    @Override
+    public boolean matches(final Path path) {
+      return true;
+    }
+
+    @Override
+    public String toString() {
+      return "**/*";
+    }
+  };
+
+  private static final WatchEvent.Kind<?>[] KINDS = {ENTRY_CREATE, ENTRY_DELETE, ENTRY_MODIFY};
+
+  private final List<PathMatcher> matchers = new ArrayList<>();
+
+  private Modifier modifier = SensitivityWatchEventModifier.HIGH;
+
+  private boolean recursive = true;
+
+  private final Class<? extends FileEventHandler> handler;
+
+  private final Path path;
+
+  private FileEventHandler handlerInstance;
+
+  FileEventOptions(final Path path, final Class<? extends FileEventHandler> handler)
+      throws IOException {
+    if (!Files.exists(path)) {
+      Files.createDirectories(path);
+    }
+    this.path = path;
+    this.handler = handler;
+    this.handlerInstance = null;
+  }
+
+  FileEventOptions(final Path path, final FileEventHandler handler) throws IOException {
+    this(path, handler.getClass());
+    this.handlerInstance = handler;
+  }
+
+  /**
+   * Append a kind filter.
+   * <p>
+   * The default filter is: {@link StandardWatchEventKinds#ENTRY_CREATE},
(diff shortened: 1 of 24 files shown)


krisztiankocsis:
Hi, I can do this today near 20:00 CET.
Are there any specific requests or things should I consider?

jknack:
Not at all, just a new module :) which I probably going to reuse in the rocker template module.

Thanks.

jknack:
Sorry, I got now what you mean. I didn't wrote anything about it yet, but for now let's make sure to use 2 spaces for indent, one statement for line, final for parameters.

I'm open to suggestion too.

jooby-filewatcher/src/main/java/org/jooby/filewatcher/FileEventOptions.java
```diff
@@ -0,0 +1,163 @@
...
+  static final PathMatcher TRUE = new PathMatcher() {
+
+    @Override
+    public boolean matches(final Path path) {
+      return true;
+    }
+
+    @Override
+    public String toString() {
+      return "**/*.*";
```
krisztiankocsis: I think it should be "**/*" since there might be files without '.'

jooby-filewatcher/src/main/java/org/jooby/filewatcher/FileEventOptions.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+  };
+
+  private static final WatchEvent.Kind<?>[] KINDS = {ENTRY_CREATE, ENTRY_DELETE, ENTRY_MODIFY };
```
krisztiankocsis: Needs a space after the opening bracket.

jooby-filewatcher/src/main/java/org/jooby/filewatcher/FileEventOptions.java
```diff
@@ -0,0 +1,163 @@
...
+  boolean recursive() {
+    return recursive;
+  }
+
+  Path path() {
+    return path;
+  }
+
+  FileEventHandler handler(
+      final Function<Class<? extends FileEventHandler>, FileEventHandler> factory) {
```
krisztiankocsis: 4 spaces used for indent instead of 2.
Is this an exception when breaking long lines in method parameters?

jooby-filewatcher/src/main/java/org/jooby/filewatcher/FileMonitor.java
```diff
@@ -0,0 +1,155 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.inject.Injector;
+
+class FileMonitor implements Runnable {
+
+  /** The logging system. */
+  private final Logger log = LoggerFactory.getLogger(getClass());
```
krisztiankocsis: Why is this not static?

jooby-filewatcher/src/main/java/org/jooby/filewatcher/FileMonitor.java
```diff
@@ -0,0 +1,155 @@
...
+  private WatchService watcher;
+
+  private final Map<WatchKey, Path> keys = new ConcurrentHashMap<>();
+
+  private final Map<Path, FileEventOptions> options = new ConcurrentHashMap<>();
+
+  private Set<FileEventOptions> optionList;
+
+  @Inject
+  public FileMonitor(final Injector injector,
```
krisztiankocsis: injector, watcher and optionList fields can be final, is there a reason why are they not final while some other non-static fields final?

jooby-filewatcher/src/main/java/org/jooby/filewatcher/FileEventOptions.java
jknack: yes, max line characters is 100, when we couldn't fit everything in one line, we use 4 spaces.

jooby-filewatcher/src/main/java/org/jooby/filewatcher/FileWatcher.java
```diff
@@ -0,0 +1,385 @@
...
+ * @author edgar
+ */
+public class FileWatcher implements Module {
+  private static final Consumer<FileEventOptions> EMPTY = it -> {
+  };
+
+  /** The logging system. */
+  private final Logger log = LoggerFactory.getLogger(getClass());
+
+  private List<CheckedFunction2<Config, Binder, FileEventOptions>> bindings = new ArrayList<>();
```
krisztiankocsis: In FileMonitor keys and options fields are both final and this field is very similar to them, do you have a rule when final should be used for fields?

jooby-filewatcher/src/main/java/org/jooby/filewatcher/FirstOfPathMatcher.java
```diff
@@ -0,0 +1,29 @@
...
+
+  private List<PathMatcher> matchers;
+
+  FirstOfPathMatcher(final List<PathMatcher> matchers) {
+    this.matchers = matchers;
+  }
+
+  @Override
+  public boolean matches(final Path path) {
+    for (PathMatcher matcher : matchers) {
```
krisztiankocsis: As I saw you like to use Java 8 stream API, whouldn't it be better to write as matchers.stream().filter(...).findFirst()... ?

jooby-filewatcher/src/main/java/org/jooby/filewatcher/FileMonitor.java
jknack: Class is a singleton, no need to use a static logger. Also, I read somewhere that static logger cause leaks within dynamic class loaders (like the one we use with jooby:run), so unless is strictly necessary I alway keep log as final instance field.

jknack:
@krisztiankocsis your comments are all in.. I moved to IDEA so there are some unrelated and required pom changes to make IDEA happy.

Thanks

PR #634, opened 2017-02-08
Flyway is not instantiated by using Named annotation fix #623

### LINKED ISSUE #623: Flyway is not instantiated by using Named annotation
It's not possible to use different instances of Flyway, since the names specified in the following code:

```
{
...use(new Flywaydb("flyway.db1");
...use(new Flywaydb("flyway.db2");
}
```

are always ignored and Jooby (or rather Guice) complains about multiple bindings of Flyway:

`A binding to org.flywaydb.core.Flyway was already configured at org.jooby.flyway.Flywaydb.configure(Flywaydb.java:183).`

jknack:


jknack:
@tryIO can you review it?

Thanks.

tryIO:
Seems good! Just checked @krisztiankocsis' changes.

PR #636, opened 2017-02-09
WebSockets: #isOpen fix #574

### LINKED ISSUE #574: WebSockets: #isOpen
There is no possibility to ask the WebSocket whether it is open.
Could you delegate that method?

org.jooby.spi.NativeWebSocket#isOpen

jknack:
* synchronized ws access

### DIFF
--- a/coverage-report/src/test/java/org/jooby/issues/Issue636.java
+++ b/coverage-report/src/test/java/org/jooby/issues/Issue636.java
@@ -1,0 +1,63 @@
+package org.jooby.issues;
+
+import java.nio.channels.ClosedChannelException;
+import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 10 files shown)


jknack:
@jschneider can you review too?

Thanks

jschneider:
Will take a look. I hope I will find time in the next 24 hours.

jschneider:
Looks good on the first glance.

Two things:
- "synchronized(this)": Is this how you are synchronizing in jooby? I'd prefer a lock, but of course "synchronized" should work.

- What happens if somebody writes into a closed web socket? 


I will take a closer look later.

jknack:
Updated. I kept sync and now throw an `Err(1001=NORMAL close)` on closed web sockets

Issue #630, opened 2017-02-07
jdbc/hikari: rename properties for multi-database usage


jknack:
Advanced hikari configuration is available via `hikari[.dbname.].property`. For example:

```
db.spock.url = "jdbc:mysql://localhost:3306/spock?createDatabaseIfNotExist=true&useSSL=false&serverTimezone=Europe/Rome"
db.spock.user = "root"
db.spock.password = "password"

hikari.spock.maximumPoolSize = 100

db.vulcan.url = "jdbc:mysql://localhost:3306/vulcan?createDatabaseIfNotExist=true&useSSL=false&serverTimezone=Europe/Rome"
db.vulcan.user = "root"
db.vulcan.password = "password"
db.vulcan.cachePrepStmts = true

hikari.vulcan.maximumPoolSize = 50
```

Here we define two databases: `db.spock` and `db.vulcan`. Then we customize poolsize for each of them via `hikari.spock.maximumPoolSize` and `hikari.vulcan.maximumPoolSize`

Due to how it works [config library](https://github.com/typesafehub/config), this syntax `hikari[.dbname].property` breaks/fails at startup time when multiple database are available.

Solution is to prefix hikari properties with database name, like: `db.spock.hikari.maximumPoolSize` or `db.vulcan.hikari.maximumPoolSize`

PR #640, opened 2017-02-12
jdbc/hikari & flywaydb: rename properties for multi-database usage fi…

jknack:
…x #630


jknack:
@tryIO Same here, appreciate your feedback.

PR #644, opened 2017-02-13
server side include

paul-hammant:


ph-ms:
Any thoughts on this functionality, Edgar? - Paul H

jknack:
Hi Paul,

 I'm coming back from holidays... and yes I'm planning to add your SSI handler in next release (which think will be next week)

:)

ph-ms:
You may well not like it when you get into the detail. But I'm sure you've a better (cleaner) way of doing the same thing. That's normally the way if my OSS contributions :)

Issue #674, opened 2017-03-18
HttpServletRequest Attributes not available to jooby App


andrenpaes:
We're trying to use jooby inside a Servlet container. In that setup, we also have a Filter that adds request scoped info into each request (using [`HttpServletRequest#setAttribute`](http://docs.oracle.com/javaee/7/api/javax/servlet/ServletRequest.html#setAttribute-java.lang.String-java.lang.Object-)).

Unfortunately, there's no way to access those attributes using jooby's `Request` object. We have an ugly workaround in place, using a ThreadLocal and a jooby before filter:
```java
before((req, rsp) -> {
    req.set("attributeName", someThreadLocal.get());
});
```

A possible solution, would be merging the `HttpServletRequest#attributes` into jooby's `locals` attributes. I have a patch that implements that, I can open a PR if it seems like a good enough solution. Let me know.

jknack:
Sounds very good. Please send your patch.

Thanks

PR #676, opened 2017-03-19
Exposes Servlet Request attributes

andrenpaes:
Fix for #674

I couldn't really find a good place to write a test to the `HttpHandlerImpl` change, but since it's a small, simple change, maybe we're good without a test.

Anyway, if you think it's worth it I can write the test, I'd just need a pointer to specific test case I could use.

### DIFF
--- a/jooby-servlet/src/main/java/org/jooby/servlet/ServletServletRequest.java
+++ b/jooby-servlet/src/main/java/org/jooby/servlet/ServletServletRequest.java
@@ -23,11 +23,7 @@ import static java.util.Objects.requireNonNull;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URLDecoder;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Enumeration;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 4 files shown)


jooby-servlet/src/main/java/org/jooby/servlet/ServletServletRequest.java
```diff
@@ -120,6 +116,15 @@ public String rawPath() {
   }
 
   @Override
+  public Map<String, Object> attributes() {
+    final Enumeration<String> attributeNames = req.getAttributeNames();
+    if (!attributeNames.hasMoreElements()) {
```
jknack: Test the empty branch too

jooby-servlet/src/main/java/org/jooby/servlet/ServletServletRequest.java
```diff
@@ -120,6 +116,15 @@ public String rawPath() {
...
+      return Collections.emptyMap();
+    }
+    return Collections.list(attributeNames).stream().collect(Collectors.toMap(name -> name, name -> req.getAttribute(name)));
```
jknack: split this line and leave one method call per line. Also for `name -> name`, use `Function.identity()`

PR #680, opened 2017-03-20
Send message to all connected web socket clients fix #645

### LINKED ISSUE #645: Send message to all connected web socket clients
I'm using Jooby's MVC routes for an API. I have also set up a websocket, to which a few clients connect. What I'm trying to do is send a message to all connected websocket clients whenever a specific http request is received in the server. This is how my route method looks like:

```
    @Path("/player")
    @Produces("application/json")
    public class PlayerRoute {
        @POST
        public Result newPlayer(Request req, @Body Player player) {
            //this is what I'm trying to achieve..
            allWebsocketSessions.foreach(session ->
                    session.send("a new player has been created")
            );
            return Results.ok();
        }
}
```

I've read jooby's documentation but can't figure out how to do it.

Thanks in advance.

jknack:


jooby/src/main/java/org/jooby/WebSocket.java
```diff
@@ -734,4 +734,56 @@ default void send(final Object data, final OnError err) throws Exception {
...
+   * Send data to all connected sessions.
+   *
+   * If the web socket is closed this method throw an {@link Err} with {@link #NORMAL} close status.
+   *
+   * @param data Data to send.
+   * @param success A success callback.
+   * @param err An err callback.
+   * @throws Exception If something goes wrong.
+   */
+  void broadcast(final Object data, final SuccessCallback success, final OnError err)
```
krisztiankocsis: Final on interface method that are not default makes no sense since they do not affect the actual implementation.

jooby/src/main/java/org/jooby/internal/WebSocketImpl.java
```diff
@@ -143,6 +152,14 @@ public boolean isOpen() {
   }
 
   @Override
+  public void broadcast(final Object data, final SuccessCallback success, final OnError err)
+      throws Exception {
+    for (WebSocket ws : sessions) {
```
krisztiankocsis: The broadcast terminates on the first error. Wouldn't it be better to try to send to all endpoint, log the exception and throw a new one if there was at least one failure?

jknack:
@krisztiankocsis thank you, I update the pull with your comments. You're right about broadcast error, but I didn't rethrow anything. Instead, I used the error callback to notify about a possible error.

Issue #731, opened 2017-04-19
IllegalReferenceCountException: refCnt: 0 (Exec + Jackson)


pierredavidbelanger:
I am trying to follow the doc to [setup thread pool](http://jooby.org/doc/async/).

I get an error while trying to parse the JSON body in a deferred handler.

Here is a code snippet that demonstrate the problem:

```java
public class App extends Jooby {

    private static class MyValueObject {
        private String field;
        public String getField() { return field; }
        public void setField(String field) { this.field = field; }
    }

    {
        use(new Exec());
        use(new Jackson());

        executor("worker1", Executors.newFixedThreadPool(10));

        post("/myvalue", deferred("worker1", (request) -> {
            MyValueObject myValueObject = request.body(MyValueObject.class);
            Thread.sleep(1000); // I am really busy here!
            return myValueObject;
        }));

        post("/myvalue2", (request) -> {
            MyValueObject myValueObject = request.body(MyValueObject.class);
            Thread.sleep(1000); // I am blocking a server worker thread
            return myValueObject;
        });

        get("/", () -> "Hello World!");
    }

    public static void main(final String[] args) {
        run(App::new, args);
    }
}
```

When I issue this `curl` command:

```bash
curl -v -X POST \
    -H 'Content-type: application/json' \
    -d '{"field": "a value"}' \
    http://localhost:8080/myvalue
```

I see this stack trace in the app console output:

```txt
[2017-04-19 15:06:14,518]-[pool-20-thread-2] DEBUG org.jooby.spi.HttpHandler - execution of: POST/myvalue resulted in exception
io.netty.util.IllegalReferenceCountException: refCnt: 0
	at io.netty.buffer.AbstractByteBuf.ensureAccessible(AbstractByteBuf.java:1408)
	at io.netty.buffer.AbstractByteBuf.checkReadableBytes0(AbstractByteBuf.java:1394)
	at io.netty.buffer.AbstractByteBuf.checkReadableBytes(AbstractByteBuf.java:1383)
	at io.netty.buffer.AbstractByteBuf.readBytes(AbstractByteBuf.java:850)
	at io.netty.buffer.CompositeByteBuf.readBytes(CompositeByteBuf.java:1798)
	at io.netty.buffer.CompositeByteBuf.readBytes(CompositeByteBuf.java:44)
	at io.netty.buffer.ByteBufInputStream.read(ByteBufInputStream.java:179)
	at java.io.InputStream.read(InputStream.java:101)
[...]

PR #735, opened 2017-04-21
A minimal test (failing) to show that a Netty buffer is not accessible

pierredavidbelanger:
outside a Netty thread (related to #731)

### DIFF
--- a/coverage-report/src/test/java/org/jooby/issues/Issue731.java
+++ b/coverage-report/src/test/java/org/jooby/issues/Issue731.java
@@ -1,0 +1,26 @@
+package org.jooby.issues;
+
+import org.jooby.test.ServerFeature;
+import org.junit.Test;
+
+import java.util.concurrent.Executors;
+
+public class Issue731 extends ServerFeature {
+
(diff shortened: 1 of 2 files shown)


pierredavidbelanger:
The last commit seems to fix the Netty buffer reference count.

I simply `.retain()` the `((ByteBufHolder) req).content()` before entering into the executor thread, and finally `.release()` it after the execution, effectively leaving the Netty buffer accessible during the deferred request.

pierredavidbelanger:
Though, I think this PR needs review from someone that knows Netty better than me, because, I do not know if it is allowed (read: does not create a leak) when I [`.release()` the buffer from a non Netty thread](https://github.com/jooby-project/jooby/pull/735/commits/e69d1babee46bd78e9229c25dd23c31411fd0344#diff-5f2b5a4d9027abfcdf86faa66feed5cbR253).

jknack:
Thank you @pierredavidbelanger what you fixed it :)
I will merge and cleanup later.
Thanks again.

Issue #718, opened 2017-04-11
Use jooby spec with ebean enhanced classes


ajcamilo:
I'm building a project that uses jooby with ebean.

I have the following entity:

```java
@Entity
public class Author {
    @Id
    public int id;
    public String name;
    public String email;
    public Author(int id, String name, String email) {
        this.id = id;
        this.name = name;
        this.email = email;
    }
    public int getId() { return id; }
    public String getName() { return name; }
    public String getEmail() { return email; }
}
```

And the following route:

```java
use("api/author")
  /**
    * Get list of all authors
    *
    * @return Returns list of all authors
    */
  .get(req -> {
    EbeanServer ebean = require(EbeanServer.class);
    List<AuthorImpl> authors = new LinkedList<>(new QAuthorImpl(ebean).findList());
    return authors;
  });
```

When I try to generate the spec with `jooby:spec` and I go inspect the RAML spec file, I see that my `Author` class has pulled all the things that ebean enhanced with.

Is the `jooby:spec` looking into the enhanced `Author` class?

jknack:
Well, yea. Guess ebean rewrite Author.class

What if you run jooby:spec in an early phase than ebean enhancer?

ajcamilo:
I don't know how I can do this. Here is the build section of my `pom.xml`

```xml
  <build>
    <plugins>
      <!-- Build fat jar -->
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-shade-plugin</artifactId>
      </plugin>

      <!-- api spec -->
      <plugin>
        <groupId>org.jooby</groupId>
        <artifactId>jooby-maven-plugin</artifactId>
        <executions>
          <execution>
            <goals>
              <goal>spec</goal>
            </goals>
          </execution>
        </executions>
      </plugin>

      <!-- data -->
      <plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>build-helper-maven-plugin</artifactId>
        <executions>
          <execution>
            <id>add-ebeanquery-source</id>
            <phase>generate-sources</phase>
            <goals>
              <goal>add-source</goal>
            </goals>
            <configuration>
              <sources>
                <source>target/generated-sources/annotations</source>
              </sources>
            </configuration>
          </execution>
        </executions>
      </plugin>
      <plugin>
        <groupId>io.repaint.maven</groupId>
        <artifactId>tiles-maven-plugin</artifactId>
        <version>2.8</version>
[...]

PR #753, opened 2017-05-02
jooby raml ignore class hidden properties

ajcamilo:
Jooby RamlType ignore properties starting with `_`, for example `_ebeanServer`.

This resolves the issue #718


ajcamilo:
@jknack if you want I can try to make this "ignore" optional through config.

jknack:
@ajcamilo Thank you for your pull and just link from general issue to this particular problem.

We need to do something similar for Swagger (guess).

ajcamilo:
I tried swagger module, and it works fine. The problem is with the RAML module.

ajcamilo:
@jknack are you delaying this on purpose? Because of the refactor the spec modules?

I'm asking this because I'm using my forked version of jooby because of this PR and another that I will make because of POJOs with UUIDs. And would like to use the official "jooby".

jknack:
Merged it. Going to release 1.1.2 by the end of this week (or early next).

Thank you, don't forget to star/follow the project

PR #770, opened 2017-05-15
add UUID type to raml spec

ajcamilo:
When RAML spec builder tries to parse a POJO that has a property of type UUID, it tries to parse like a normal class. 

But the output should be something like this:

```
types:
  UUID:
    type: string
    pattern: ^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$
```
(reference: https://github.com/raml-org/raml-spec/issues/483#issuecomment-219048271)

This is what this PR resolves for the `java.util.UUID`.

### DIFF
--- a/jooby-raml/src/main/java/org/jooby/internal/raml/RamlType.java
+++ b/jooby-raml/src/main/java/org/jooby/internal/raml/RamlType.java
@@ -32,6 +32,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.UUID;
 
 import com.google.common.collect.ImmutableList;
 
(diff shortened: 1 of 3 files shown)


jooby-raml/src/main/java/org/jooby/internal/raml/RamlType.java
```diff
@@ -196,6 +204,10 @@ private static RamlType simpleParse(final Type type) {
         enums.add(((Enum) value).name());
       }
       complex.values = enums;
+    } else if(UUID.class.isAssignableFrom(rawType)){
```
jknack: Add space after `if` before closing `{`

PR #796, opened 2017-05-25
support neo4j driver for jooby

sbcd90:
This PR is created to propose the addition of `neo4j` graph database driver for jooby.
The PR contains the implementation of 

- neo4j driver for jooby.
- neo4j session store for jooby.

The test cases & documentation is also included with this PR.


jooby-neo4j/src/main/java/org/jooby/neo4j/Neo4j.java
```diff
@@ -0,0 +1,259 @@
...
+   * @return This module
+   */
+  public Neo4j properties(final BiConsumer<Properties, Config> properties) {
+    this.properties = requireNonNull(properties, "properties callback is required");
+    return this;
+  }
+
+  @Override
+  public Config config() {
+    return ConfigFactory.parseResources(Neo4j.class, "org.jooby.neo4j/neo4j.conf");
```
jknack: does it work? It should be: `ConfigFactory.parseResources(Neo4j.class, "neo4j.conf");`

jooby-neo4j/src/main/java/org/jooby/neo4j/Neo4j.java
```diff
@@ -0,0 +1,259 @@
...
+
+  private Properties options(final Config config) {
+    Properties properties = new Properties();
+
+    properties.put(DBProperties.SERVER_ROOT_URI, config.getString("uri"));
+    if (config.hasPath("arrayBlockSize")) {
+      properties.put(DBProperties.ARRAY_BLOCK_SIZE, config.getString("arrayBlockSize"));
+    }
+    properties.put(DBProperties.DATABASE_DIR,
+      config.hasPath("databaseDir") ? config.getString("databaseDir"): System.getProperty("java.io.tmpdir"));
```
jknack: remove: `System.getProperty("java.io.tmpdir")` with `config.getString("application.tmpdir")`

jooby-neo4j/src/main/java/org/jooby/neo4j/Neo4j.java
```diff
@@ -0,0 +1,259 @@
...
+  @Override
+  public Config config() {
+    return ConfigFactory.parseResources(Neo4j.class, "org.jooby.neo4j/neo4j.conf");
+  }
+
+  private Properties options(final Config config) {
+    Properties properties = new Properties();
+
+    properties.put(DBProperties.SERVER_ROOT_URI, config.getString("uri"));
+    if (config.hasPath("arrayBlockSize")) {
```
jknack: It is better to dump all the properties from config into the properties object. It simplifies the code (no ifs) but also it is more flexible

jooby-neo4j/src/main/java/org/jooby/neo4j/Neo4j.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    return properties;
+  }
+
+  private Config neo4j(final Config config) {
+    Config $neo4j = config.getConfig("neo4j");
+
+    if (config.hasPath("neo4j." + db)) {
+      $neo4j = config.getConfig("neo4j." + db);//.withFallback($neo4j);
```
jknack: we usually fallback to default configuration, why did you comment the fallback?

jooby-neo4j/src/main/java/org/jooby/neo4j/Neo4jSessionStore.java
```diff
@@ -0,0 +1,255 @@
...
+  private String decode(final String key) {
+    String value = key;
+    if (value.charAt(0) == UDOLLAR) {
+      value = DOLLAR + value.substring(1);
+    }
+    return value.replace(UDOT, DOT);
+  }
+
+  private static Pair<GraphDatabaseService, GraphAwareRuntime> getOrCreateGraph(String databaseDir) {
+    if (dbService == null && graphRuntime == null) {
```
jknack: Feel we should export dbService and graphRuntime from Ne4j module. For example who/when all these services go down?
This two services should be exposes by Neo4j and injected at construction time here in Neo4jSessionStore

sbcd90: Hi @jknack , The `dbService` and `graphRuntime` cannot be exposed from Neo4j module because neo4j module should be `used to access a remote neo4j db` while the Neo4j session store creates a local neo4j db & tries to access it using the `dbService` and `graphRuntime` objects.

The test cases demonstrates how they should be used.

sbcd90:
Hi @jknack ,

Thanks for your review & comments. I have fixed most of them.
Plz have a look.

jooby-neo4j/src/main/java/org/jooby/neo4j/Neo4jSessionStore.java
jknack: Got it, but won't be useful too:

- Add support for local db too Neo4j
- Add support for remote db to Neo4jSessionStore

We can probably merge this and then add that in a new pull request.

sbcd90: sure, I'll add support for them in a new PR.

jknack: Well kind of, we still need to figure it out how to shutdown the local instance created by Neo4jSessionStore

sbcd90: So, in case of ehcache store, an instance of `EhCache` is taken as input. So, do you want to follow that approach & take `dbService` and `graphRuntime` as inputs?

jknack: Right. Something similar to that, but still we need to manage the lifecycle of those objects (like ehcache module does).

Will be nice to have something like:

```
db = fs
```
Or `local` and then if we do:

```
{
  use(new Neo4j());
}
```

The module exposes those services...but don't like is that module will exposes completely different services BoltDBAccess vs DbService/GraphRuntime :S

It is probably better to write a LocalNeo4j? or EmbeddedNeo4j module?

Thoughts?

sbcd90: I think yes. Lets first consider the embedded neo4j..I'll change the neo4j module to provide the 
`dbService` & `graphRuntime` objects.
Plz let me know if you agree..

jknack: Let;s create a new module: EmbeddedNeo4j and leave the existing Neo4j module as it is now.

sbcd90: Ok..I'll create a new PR for the EmbeddedNeo4j module & leave this PR untocuhed..is it ok for you?

jknack: Sure, np.

jknack:
I changed the project structure and your pull can't be merge it, sorry for that.

But don't worry going to pick your commit and merge them my self later (unless you want to do it :)

sbcd90:
Hi @jknack ,

I can make the changes myself..however, do you want me to merge my 2 PRs into 1 module?
Plz let me know

jknack:
That will be awesome!

sbcd90:
Hi @jknack , I have merged the modules & updated the PR. Please let me know if this PR can be merged now..

PR #800, opened 2017-05-27
remove force recompilation of classes from jooby:spec

ajcamilo:
I was having issues with my build cycle and the main reason was because `jooby:spec` forks the compilation to generate the spec file and after this the default compilation occurs.

`jooby:spec` executes the build to the `process-classes` phase.

Besides my compilation problems (mainly because of ebean and querybean enhacement), another down side of this is that the compilation is much slower (2 compilations).

This PR just removes the execution of the `process-classes` and if the plugin is set to execute in the phase `process-classes` everything works as expected.


jknack:
I changed the project structure and your pull can't be merge it, sorry for that.

But don't worry going to pick your commit and merge them my self later (unless you want to do it :)

ajcamilo:
rebase done :wink:

jknack:
Merged it. Much appreciated!

Issue #805, opened 2017-06-01
RequestLogger with query params


DrorBuhnik:
Currently RequestLogger log url path only without query params
I look but didn't find if NCSA format define what exactly needs to go into the request part of the log

Can we add query params to the log?
I can create pull request with the change (If I do, do you want it to be controllability, so the users can turn it on or off like they config the data format?)

jknack:
Pull request with ability to turn on (off by default) will be great:

```
use(new RequestLogger().queryString());
```

PR #808, opened 2017-06-03
Add option to log url request including query string

DrorBuhnik:
PR for #805

### DIFF
--- a/jooby/src/main/java/org/jooby/RequestLogger.java
+++ b/jooby/src/main/java/org/jooby/RequestLogger.java
@@ -197,6 +197,7 @@ public class RequestLogger implements Route.Handler {
   private static final char BL = '[';
   private static final char BR = ']';
   private static final char Q = '\"';
+  private static final char QUERY = '?';
 
   private static Function<Request, String> ANNON = req -> DASH;
 
@@ -210,6 +211,8 @@ public class RequestLogger implements Route.Handler {
   private Function<Long, String> df;
 
   private boolean latency;
+
+  private boolean queryString;
 
   private boolean extended;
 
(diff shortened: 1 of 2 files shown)


jknack:
lovely! but we need a test :)

See [RequestLoggerTest](https://github.com/jooby-project/jooby/blob/72a7bd0b1cc2c7b92f8bd2717b529206f4c7b1f0/jooby/src/test/java/org/jooby/RequestLoggerTest.java)

jooby/src/test/java/org/jooby/RequestLoggerTest.java
```diff
@@ -147,6 +169,13 @@ private Block path(final String path) {
     };
   }
 
+  private Block query(final String query) {
+    return unit -> {
+      Request req = unit.get(Request.class);
+      expect(req.queryString()).andReturn(Optional.of(query)).atLeastOnce();
```
DrorBuhnik: I'm not sure I like it that I'm the only one that uses .atLeastOnce() in the test
Any better way? (The RequestLogger call several times to queryString())

jknack: You could move the queryString() as part of the if block ... but I'm OK with this.

jooby/src/main/java/org/jooby/RequestLogger.java
```diff
@@ -250,9 +250,8 @@ public void handle(final Request req, final Response rsp) throws Throwable {
       sb.append(Q).append(req.method());
       sb.append(SP);
       sb.append(req.path());
-      if (queryString && req.queryString().isPresent()) {
-        sb.append(QUERY);
-        sb.append(req.queryString().get());
+      if (queryString) {
+        req.queryString().ifPresent(s -> sb.append(QUERY).append(s));
```
DrorBuhnik: I refactored into single queryString() call as you suggested
Better?

Issue #814, opened 2017-06-12
kotlin: make Kooby class open 


jknack:
So we can extend if need it:

```
class App: Kooby() {
  ...
}
```

Kooby is an specialised and kotlin version of `org.jooby.Jooby`

PR #815, opened 2017-06-13
Makes Kooby Class open

vjames19:
#814 

First of all I want to thank you for this amazing framework!

I'm using it with Kotlin and just stumbled upon the need to extend the Kooby class to be able to use it in tests.


jknack:
Thanks!

Can you please add an [example](https://github.com/jooby-project/jooby/tree/master/modules/jooby-lang-kotlin/src/test/kotlin/examples/kotlin) where we extend Kooby?

vjames19:
@jknack I changed the App.kt file to extend Kooby instead of Jooby.

vjames19:
@jknack Is there anything else I should add to this one?

Issue #809, opened 2017-06-04
kotlin: param, header, body idioms


jknack:
Today:

```kt
get {
  val str = param("name").value
  val intval = param("name").intValue
  val type =  param("x").to(MyType::class)
}
```

Proposed new idioms using reified types:

```kt
get {
  val str = param<String>("name")
  val intval = param<Int>("name")
  val type =  param<MyType>("x")
}
```

PR #816, opened 2017-06-13
using reified types to extract param, header and body.

vjames19:
#809 

Let me know if you want the rest of the Mutant extension functions to also use reified types.

### DIFF
--- a/modules/jooby-lang-kotlin/src/main/kotlin/org/jooby/Jooby.kt
+++ b/modules/jooby-lang-kotlin/src/main/kotlin/org/jooby/Jooby.kt
@@ -305,12 +305,24 @@ val Mutant.doubleValue: Double
 val Mutant.value: String
   get() = value()
 
+inline fun <reified T> Mutant.to(): T {
+  return this.to(T::class.java)
+}
+
 fun <T:Any> Mutant.to(type: KClass<T>): T {
   return this.to(type.java)
 }
 
+inline fun <reified T> Mutant.to(contentType: MediaType): T {
+  return this.to(T::class.java, contentType)
+}
+
 fun <T:Any> Mutant.to(type: KClass<T>, contentType: MediaType): T {
   return this.to(type.java, contentType)
+}
+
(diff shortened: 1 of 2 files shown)


modules/jooby-lang-kotlin/src/main/kotlin/org/jooby/Jooby.kt
```diff
@@ -305,14 +305,26 @@ val Mutant.doubleValue: Double
 val Mutant.value: String
   get() = value()
 
+inline fun <reified T> Mutant.to(): T {
```
jknack: lovely! I forgot about these `.to` methods.

jknack:
Same here, let's add a couple of example for each using the reified version and normal/java version of them

vjames19:
@jknack I added some examples on the last commit. Is this where you want them https://github.com/vjames19/jooby/blob/ce5db1816a6f3a7733059e4b7ab34aabc5903add/modules/jooby-lang-kotlin/src/test/kotlin/examples/kotlin/Fn.kt?

jknack:
Perfect! Also I usually prefer one commit for pull request (unless the issue is extremely complex), so once review comments are in, I amend the commit and then force push.

vjames19:
@jknack Or you can also click the dropdown on "Merge Pull Request" button and select "Squash and merge" this would let you write a commit message and it would only show up as one commit on the master branch.

You can also make this the default behavior for PRs in this repo, so that you don't forget about it when you merge.

Here is the official doc: https://github.com/blog/2141-squash-your-commits

Here I found some other project discussing about the topic: https://github.com/SatelliteQE/robottelo/issues/3803

This would remove the "one commit" requirement and would keep the master branch commit history clean.

What do you think?

jknack:
@vjames19 didn't know about this, sounds perfect!

jknack:
love it: https://github.com/jooby-project/jooby/commit/f2df155cbad52d4c66f9b958f32a1c2578c18de7

Issue #761, opened 2017-05-08
Gradle: Hotswap for Kotlin code does not work


nlochschmidt:
I have a gradle based jooby project with most of the app code written in Kotlin. It works quite well however unfortunately the Hotswap feature `joobyRun` is not working.

Versions: Jooby 1.1.1, Gradle 3.4, Kotlin 1.1.2-3, Java 8u121

Steps to reproduce:
1. Check out my fork of the [kotlin-starter](https://github.com/nlochschmidt/kotlin-starter) project
2. Start the jooby app with `./gradlew joobyRun`
3. Go to http://localhost:8080. You should see `Hello Kotlin!`
4. Change response of the get handler in App.kt to `"Good day $name!"`
5. Reload the page

At this point the page should still display `Hello Kotlin!` instead of `Good day Kotlin!`

Workarounds:
1. Restart the jooby instance between each change (slow and cumbersome)
2. Start a second gradle process with `./gradlew compileJava -t` to continuously recompile the code. This seems to work but I have seen race conditions with the class loader when the reload happens before the recompilation is finished.

I am also a bit confused by the log output. Why is there an entry for `**/*.class` instead of `**/*.java` and `**/*.kt`?

```
>>> jooby:run[info|Daemon worker]:   includes: [**/*.class:**/*.conf:**/*.properties:*.js:src/*.js]
>>> jooby:run[info|Daemon worker]:   excludes: []
```

jknack:
I'm not a gradle user and don't know how to fire compilation from a single gradle process :(

> I am also a bit confused by the log output. Why is there an entry for **/*.class instead of **/*.java and **/*.kt?

In my experience [incremental] compilation works better from IDE. Ideally when you save the file, compilation takes place. IDE regenerates .class file and Jooby reload your application.

nlochschmidt:
Thanks for the quick response. Yes, using the IDE to compile classes is an option that works, however imo there is value in a solution that's independent from a particular editor or IDE.

I think I narrowed down the relevant places in the gradle script that would need to change. First of all the Kotlin extension `.kt` should be added here: [JoobyTask.java#L102](https://github.com/jooby-project/jooby/blob/v1.1.1/jooby-gradle-plugin/src/main/java/org/jooby/run/JoobyTask.java#L102), right? 

That change alone allows a new workaround where I can manually add the Kotlin directory as java src dir to my projects build.gradle file like so:

```
sourceSets.main.java {
    srcDirs "src/main/kotlin"
}
[...]

PR #820, opened 2017-06-19
Gradle Plugin: Towards hot-reloading for Kotlin

nlochschmidt:
Related to #761 

This should be seen as a first step towards full Kotlin hot-reload support in Gradle, but it already works when the Kotlin sources are alongside the Java sources under `/src/main/java`. According to the Kotlin Gradle plugin docs it is a valid location for Kotlin sources: https://kotlinlang.org/docs/reference/using-gradle.html#android-studio

It's certainly better than no hot-reloading at all 😉


jknack:
does compilation works once you changed a .kt file?

nlochschmidt:
Yes, as long as the the .kt file is in the Java source folder.

jknack:
What version of gradle? Remember it didn't work before

nlochschmidt:
Uh, ok. I tried it using my fork of the kotlin-starter that uses gradle version 3.4. You can get the configuration I used for testing from https://github.com/nlochschmidt/kotlin-starter/tree/auto-reload-test (Branch auto-reload-test)

nlochschmidt:
I forgot to create a branch on my fork of jooby before opening the pull request. I now needed to sync it with upstream so the PR got closed. In any case there wasn't much code here anyway, just this: https://github.com/nlochschmidt/jooby/commit/0b9d9861f7c5d5e483151ec9723048bf7051ace0

PR #855, opened 2017-08-16
upgrade jetty to v9.4.6

paul-hammant:


paul-hammant:
Thoughts @jnack ?

jknack:
version bump isn't enough :(

need a closer look to see why failing

paul-hammant:
I'm going to trot out that "it worked for me" claim, then hang my head in shame ....

paul-hammant:
I've just pulled all your commits to my workstation and the build still passes for me.

PR #859, opened 2017-08-29
Making the src extensions that trigger rebuild configurable

sphogan:
This will allow people to configure the src extensions that trigger rebuild so that they can add Kotlin or Rocker file extensions to trigger re-compilation.

### DIFF
--- a/doc/doc/maven-plugin/maven.md
+++ b/doc/doc/maven-plugin/maven.md
@@ -56,6 +56,7 @@ It's worth to mention that dynamic reload of classes is done via {{jboss-modules
     </includes>
     <excludes>
     </excludes>
+    <srcExtensions>.java,.kt,.conf,.properties</srcExtensions>
   </configuration>
 </plugin>
 ```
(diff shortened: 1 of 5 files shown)


jknack:
Can you please document the new property for maven and gradle?

You will find documentation inside the `/doc` folder as markdown files

Thanks

jknack:
Oh and can you please add `.kt` as default lookup extension too?

sphogan:
I've added documentation, made `.kt` a default lookup extension, and altered `JoobyMojo` so that it will do the same XML footwork that `JoobyTask` is doing.

I had tested it with Gradle and Kotlin and with Maven and Rocker, but I hadn't tested Maven with Kotlin and there was an issue with the classpath on subsequent runs that would cause re-compilation to fail.

Let me know if you want any other changes and thanks for a truly wonderful project.

jknack:
@sphogan can you rebase and update? (sorry, for taking so long to merge it)

sphogan:
I've rebased it onto the current master.  Thanks for the excellent project!

PR #876, opened 2017-09-26
847 - Added implementing class directly to MvcHandler

derKrischan:
... instead of Method.getDeclaringClass() to avoid abstract class instantiation attempts.
Fixed some compiler warnings in modified classes.

### DIFF
--- a/jooby/src/main/java/org/jooby/Route.java
+++ b/jooby/src/main/java/org/jooby/Route.java
@@ -1965,13 +1965,14 @@ public interface Route {
 
   /**
    * A handler for a MVC route, it extends {@link Handler} by adding a reference to the method
-   * behind this route.
+   * and class behind this route.
    *
    * @author edgar
    * @since 0.6.2
    */
   interface MethodHandler extends Handler {
     Method method();
+    Class<?> implementingClass();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


jooby/src/main/java/org/jooby/Route.java
```diff
@@ -1965,13 +1965,14 @@ default void handle(final Request req, final Response rsp, final Route.Chain cha
...
    * A handler for a MVC route, it extends {@link Handler} by adding a reference to the method
-   * behind this route.
+   * and class behind this route.
    *
    * @author edgar
    * @since 0.6.2
    */
   interface MethodHandler extends Handler {
     Method method();
+    Class<?> implementingClass();
```
jknack: let's use the existing `declaringClass`

jknack:
Looks good, but let's not introduce a new method (implementingClass), declaringClass is enough.

jooby/src/main/java/org/jooby/Route.java
derKrischan: OK. I will rename it to declaringClass, but I'm not sure what you mean by "the existing". Method.getDeclaringClass() exists but we can't use it because it points to the declaring abstract class. So I guess you just want m to rename implementingClass() into declaringClass()?

jknack: You're right. I confused Handler with Route.Definition class

derKrischan: Then I suggest to leave it as "implemetingClass" because it is what the name says. In 99% it will also be the declaring class but not in case of abstract classes.

jknack: Ok, yea. Let's do it

PR #878, opened 2017-09-27
Change default behavior of Request.locale()

### LINKED ISSUE #877: Default behavior of Req.locale() should be Locale.lookup()
> I have asked about this on Gitter before and promised to create an issue but didn't get around to it until now.

In my opionion, the current behavior of `Req.locale()` ist wrong. E.g. there is currently this test:
https://github.com/jooby-project/jooby/blob/8ae900fe3a3b724e40becec7bd1671785faa01ff/modules/coverage-report/src/test/java/org/jooby/issues/Issue273.java#L12-L16
https://github.com/jooby-project/jooby/blob/8ae900fe3a3b724e40becec7bd1671785faa01ff/modules/coverage-report/src/test/java/org/jooby/issues/Issue273.java#L49-L51

I think the correct response for a missing `en-AU` language should not be the default, but the closest available language, which is `en` instead of `fr-CA`. 

The concrete issue we have been running into is that we had a language list of `en, de` and the `Accept-Language` has been `de-DE` but `Req.locale()` yielded `en`, which intuitively does not make sense.

nlochschmidt:
Uses `Locale#lookup` instead of `Locale#filter` + findFirst and fallback. As a result, it will return the locale with the closest match in case that is possible.

Example: with a language setting of "en, de", a request with the `Accept-Language` header set to `de-CH` will now return `de` instead of `en`.

Closes #877

### DIFF
--- a/jooby/src/main/java/org/jooby/Request.java
+++ b/jooby/src/main/java/org/jooby/Request.java
@@ -1150,8 +1150,7 @@ public interface Request extends Registry {
    * @return A matching locale.
    */
   default Locale locale() {
-    return locale((ranges, locales) -> Locale.filter(ranges, locales).stream()
-        .findFirst()
+    return locale((priorityList, locales) -> Optional.ofNullable(Locale.lookup(priorityList, locales))
         .orElse(locales.get(0)));
   }
 
--- a/modules/coverage-report/src/test/java/org/jooby/issues/Issue273.java
+++ b/modules/coverage-report/src/test/java/org/jooby/issues/Issue273.java
@@ -48,7 +48,7 @@ public class Issue273 extends ServerFeature {
 
     request().get("/273")
         .header("Accept-Language", "en-AU")
-        .expect("fr-CA");
+        .expect("en");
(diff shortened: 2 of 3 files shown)


jknack:
Looks good, but the build is broken. Run `./build/coverage.sh` to make sure everything is good

jknack:
Also, think: `req.locale(Locale::lookup)` does what you need

nlochschmidt:
@jknack Fixed the build. I was having trouble getting it to run locally.
> Also, think: req.locale(Locale::lookup) does what you need
Not really it should fallback to the default in case there is no match which `lookup` alone does not.

Do you think the current default behavior of `Request#locale()` is the sensible default? If so why? 

Maybe have a look at the PlayFramework, because they as well do it like I proposed.

jknack:
> Do you think the current default behavior of Request#locale() is the sensible default? If so why?

No, think we do have a bug as you demonstrated in test 273. I was giving you a `workaround` until next release.

Thank you for pull

PR #881, opened 2017-10-06
Adding chain operator

### LINKED ISSUE #866: Ability to pipe use(path, jooby) calls under common prefix
Suppose we have two Jooby instances which routes should start with common prefix:

```
class Foo extends Jooby { get("/1", () -> "111" ); }
class Bar extends Jooby { get("/1", () -> "111" ); }
```

Obviously we can do somethig like this:

```
public class App extends Jooby {

    static final String API_V1 = "/api/v1";

    {
        use(API_V1 + "/foo", new Foo());
        use(API_V1 + "/bar", new Bar());
    }

    public static void main(final String[] args) {
        run(() -> new App(), args);
    }
}
```

... and then hit `/api/v1/foo/1` and `/api/v1/bar/1`.

But it would be much nicer to do something like this:

```
chain("/api/v1")
    .use("/foo", new Foo())
    .use("/bar", new Bar());
```

Actually, it can be easily achieved by writing small helper:

```
public class Routes {

    private final String routePrefix;
    private final Jooby parent;

    private Routes(String routePrefix, Jooby parent) {
        this.routePrefix = routePrefix;
        this.parent = parent;
    }

    public Routes add(String path, Jooby child) {
        parent.use(routePrefix + path, child);
        return this;
    }

    public static Routes chain(String path, Jooby parent) {
        return new Routes(path, parent);
    }
}
```

```
Routes.chain("/api/v1", this)
    .add("/foo", new Foo())
    .add("/bar", new Bar());
```

... but built-in support is always better.

paulovictorv:
Resolves #866 
@jknack I'm not sure if I tested it correctly, can you check?

### DIFF
--- a/jooby/src/main/java/org/jooby/Jooby.java
+++ b/jooby/src/main/java/org/jooby/Jooby.java
@@ -582,7 +582,7 @@ import java.util.stream.Collectors;
  */
 public class Jooby implements Router, LifeCycle, Registry {
 
-  /**
+    /**
    * <pre>{@code
    * {
    *   on("dev", () -> {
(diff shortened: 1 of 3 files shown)


jknack:
Thank you, this work and it is done properly. Just decided to do it more generic with #897

Issue #844, opened 2017-07-29
frontend module


jknack:
The [assets module](http://jooby.org/doc/assets/) works quite well today, but it is limited to:

- implementation of new processor
- keep existing processor up-to-date

A good feature of the asset module is that we don't have to deal with nodejs installations and the application just work in any environment with a simply `jooby:run`

We are going to add a frontend module using: https://github.com/eirslett/frontend-maven-plugin which allows us to keep clean and local nodejs installation but also all the variety of frontend tools available today for Javascript, including:

- webpack
- rollup
- grunt/gulp
- etc..

At build time we don't need much and the https://github.com/eirslett/frontend-maven-plugin can process and package assets results for us.

We need to extend/integrate the https://github.com/eirslett/frontend-maven-plugin at development time, specially with `jooby:run`

PR #887, opened 2017-10-10
Fix artifactId for jooby-assets-react

elaatifi:


jknack:
Thanks for this, but I think this module is going to be removed in next release (1.2.0) in favor of the new frontend module #844

PR #905, opened 2017-10-20
Enable client certificate authentication. Fix #904.

### LINKED ISSUE #904: Cannot enable client certificate authentication
I'm having trouble enabling client certificate authentication on the server side. I set the property `ssl.trust.cert` to point to a .PEM file on the classpath, and I can verify that the file is loaded (using `-Djavax.net.debug=all`). However, when I connect from a client, the server sends its own certs, but never requests certificates from the client. In the SSL debug output, it proceeds to `*** ServerHelloDone`, and never checks the client certs at all. It just moves on to complete the rest of the handshake. So, the client can use any cert, or none at all, and still connect.

Looking at `SslContextBuilder`, the `clientAuth` method is never called. Since the default value of `clientAuth` is `ClientAuth.NONE`, that's the value that gets passed down to Netty's `JdkSslContext`, and client auth is never enabled.

The fix is to call `clientAuth` in `NettySslContex.build` (line 241):
```
    if (conf.hasPath("ssl.trust.cert")) {
      scb.trustManager(toFile(conf.getString("ssl.trust.cert"), tmpdir))
         .clientAuth(ClientAuth.REQUIRE);
    }
```

There is a test for this: `NettySslContextTest.sslTrustCert`, but it doesn't check for the `clientAuth` change. To check for the failure (and verify the fix), the expect block should be (line 55):
```
        .expect(unit-> {
          SslContextBuilder scb = unit.get(SslContextBuilder.class);
          expect(scb.trustManager(Paths.get("target", "unsecure.crt").toFile())).andReturn(scb);
          expect(scb.clientAuth(ClientAuth.REQUIRE)).andReturn(scb);
        })
```

I'm running Jooby 1.1.3 with jooby-netty on Open JDK 1.8.0_144.

dogrover:
- NettySslContext now explicitly requests client auth when `ssl.trust.cert` is used.
- NettySslContextTest now checks for this setting.

### DIFF
--- a/modules/jooby-netty/src/main/java/org/jooby/internal/netty/NettySslContext.java
+++ b/modules/jooby-netty/src/main/java/org/jooby/internal/netty/NettySslContext.java
@@ -222,6 +222,7 @@ import io.netty.handler.ssl.ApplicationProtocolConfig.Protocol;
 import io.netty.handler.ssl.ApplicationProtocolConfig.SelectedListenerFailureBehavior;
 import io.netty.handler.ssl.ApplicationProtocolConfig.SelectorFailureBehavior;
 import io.netty.handler.ssl.ApplicationProtocolNames;
+import io.netty.handler.ssl.ClientAuth;
 import io.netty.handler.ssl.OpenSsl;
 import io.netty.handler.ssl.SslContext;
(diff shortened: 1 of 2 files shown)


jknack:
Fix for #905

PR #917, opened 2017-10-22
Introduce NoOpRouter

nedtwigg:
The diff for this PR is big, but it's actually a small change.  The goal of the PR is to create `NoOpRouter`, which is an implementation of `Router` which allows defining routes purely in terms of its URL scheme.  This makes refactoring routes (especially reverse routes) a lot easier.

```java
public class ArticleModule implements Jooby.Module {
  static final Route.Definition USER_TITLE = NoOpRouter.caseSensitive().get("/:user/:title");

  @Override
  public void configure(Env env, Config conf, Binder binder) throws Throwable {
    env.router().append(USER_TITLE, req -> {
      return req.params("user") + " wrote " + req.params("title");
    });
  }

  public static String urlUserTitle(String user, String title) {
    return USER_TITLE.reverse(user, title);
  }
}
```

I did this in 4 steps (each with their own commit):

1. Change the `appendDefinition()` method in Jooby from `private` to `public`, and make it part of `Router` interface.
2. That makes it possible to move almost all of Jooby's router implementation into default methods of `Router`.
3. Create `NoOpRouter`, which implements `Router` by simply returning route definitions.  This also is a succinct list of the `Router` methods which are "stateful", and can't be implemented by a `NoOpRouter`.
4. Add `Router.append(Route.Definition def, Route.Handler handler)` so that a route's URL scheme can be defined separately from its handler.

### DIFF
--- a/jooby/pom.xml
+++ b/jooby/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.jooby</groupId>
     <artifactId>jooby-project</artifactId>
-    <version>1.6.2</version>
+    <version>1.6.3-SNAPSHOT</version>
   </parent>
 
   <modelVersion>4.0.0</modelVersion>
--- a/modules/coverage-report/pom.xml
+++ b/modules/coverage-report/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jooby</groupId>
     <artifactId>modules</artifactId>
-    <version>1.6.2</version>
+    <version>1.6.3-SNAPSHOT</version>
   </parent>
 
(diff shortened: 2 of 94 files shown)


nedtwigg:
I guess this could be accomplished almost as easily with `RoutePattern`, rather than introducing `NoOpRouter`...

nedtwigg:
I think `NoOpRouter` is a bad idea, but the refactoring of parts of `Jooby` into `Router` might be useful.  `NoOpRouter` is really only useful as a quick sanity check for which parts of `Router` are stateful, and which parts are syntactic sugar that eventually get inserted into that state.  Feel free to cherry-pick none/any of this that is useful and close :)

Issue #924, opened 2017-10-26
Cannot use chain.next() with current MVC route implementation


nfleury:
Using "script" routes you can do something like that to acces Route.Chain

```
 get("/", (req,res,chain) -> {
...		    	
});
```

But when using MVC route, I was expecting something like this to work:

```
 @GET
  public void home(Reques req, Response res, Route.Chain chain) {
     chain.next(req,res)
  }
```

The Request and Response are properly injected but not Route.Chain .

jknack:
we can probably add it, but in my mind filters make sense as script/lambdas, no?

nfleury:
It does make sense, but on the other hand I really don't see why this feature wouldn't be accessible when using MVC route ? 

I chose to migrate all my routes from lambda to MVC routes because I started to have too many routes and for debugging purpose I find it better like that. 

Invoking the next route in chain can be very useful in some case, but right now I would have to write a  separated class implemeting Route.Filter that would not be part of my "MVC routes" when it's just a handler like others with only this special parameter.

jknack:
Ok, going to add.

PR #927, opened 2017-10-27
Support path when using MVC class

ajcamilo:
Now you can specify the `path` when using an MVC route.

Example:

```java
use("/pets", MyMVCRoute.class);
```


jknack:
Fix #924

PR #959, opened 2017-12-14
Allow NettyServer to run without executor

### LINKED ISSUE #706: allow to run code in the IO Thread?
We don't today. Application code always run in a worker thread.

Do you think we should allow it?

### LINKED ISSUE #956: Who is using Jooby?
I would like to know what companies and applications or who are using Jooby.

Thanks all.

sschepens:
This allows application code to run on io threads by setting executor threads to 0. Should fix #706


jknack:
It is a nice and clever solution, thank you.

I'm working in the next major release and believe is better to add IO thread execution there and leave 1.x as it is. Also, we need to find a way to implement such feature across servers (utow & jetty) under a simple and unified API.

There is something working already for 2.x so I invite you to have a look and join to the development effort (it is a complete rewrite)

PR #969, opened 2018-01-01
Setting multiple redirect option with last requested html url

S66D:
This is a temporarily fix for pac4j issue [#1074](https://github.com/pac4j/pac4j/issues/1074)

### DIFF
--- a/modules/jooby-guava-cache/src/main/java/org/jooby/guava/GuavaCache.java
+++ b/modules/jooby-guava-cache/src/main/java/org/jooby/guava/GuavaCache.java
@@ -420,16 +420,13 @@ public abstract class GuavaCache<K, V> implements Jooby.Module {
       List<TypeLiteral> types = cacheType(name, cache, getClass().getGenericSuperclass());
 
       types.forEach(type -> {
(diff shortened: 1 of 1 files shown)


modules/jooby-guava-cache/src/main/java/org/jooby/guava/GuavaCache.java
```diff
@@ -420,16 +420,13 @@ public void configure(final Env env, final Config conf, final Binder binder) {
...
+        binder.bind(Key.get(type, Names.named(name))).toInstance(cache);
         if (name.equals("cache")) {
           // unnamed cache for default cache
           binder.bind(type).toInstance(cache);
           // raw type for default cache
           binder.bind(type.getRawType()).toInstance(cache);
-        }
-        if (name.equals("session")) {
-          binder.bind(Key.get(type, Names.named(name))).toInstance(cache);
+        } else if(!name.isEmpty()) { // any type of cache names under "guava.cache" in conf file even for the "session" name
```
jknack: Did you want to commit this change?

PR #982, opened 2018-01-18
fix ebean: add query-beans maven integration

ajcamilo:
When `jooby:run` was reloading a project with ebean, entity classes where not being enhanced.


ajcamilo:
this has the problem that hot swapping is a little greedy, and starts reloading even when the project is still being compiled.

Issue #993, opened 2018-01-29
Guava version is out of date - and threre's breaking API changes


paul-hammant:
https://github.com/jooby-project/jooby/search?utf8=%E2%9C%93&q=%22guava.version%22&type=

Of course Jooby works fine.  Its other libraries that are further ahead with their dep on Guava that won't work with the older version.  This happens so often in Java-land with .. with Guava in particular :-(

Latest is  presently <version>23.6-jre</version>

paul-hammant:
24.0-jre now

feffi:
also see https://github.com/jooby-project/jooby/issues/1129

PR #994, opened 2018-01-31
upgrade guava 

paul-hammant:
Re #993 

Build w/ tests works.


paul-hammant:
And this is why you never fail the build on a coverage drop!!

jknack:
finally done here: https://github.com/jooby-project/jooby/commit/00892a49d4ca8e4f3bf97cd635ddfd5fa1005ff7

paul-hammant:
Yay!

PR #1007, opened 2018-02-07
upgrade jackson

paul-hammant:
all tests pass

### DIFF
--- a/modules/jooby-bom/pom.xml
+++ b/modules/jooby-bom/pom.xml
@@ -12,7 +12,7 @@
   <properties>
   <jooby.version>1.2.3</jooby.version>
   <funzy.version>0.1.0</funzy.version>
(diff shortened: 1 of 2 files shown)


modules/jooby-bom/pom.xml
```diff
@@ -12,7 +12,7 @@
   <properties>
   <jooby.version>1.2.3</jooby.version>
   <funzy.version>0.1.0</funzy.version>
-  <jackson.version>2.9.2</jackson.version>
+  <jackson.version>2.9.4</jackson.version>
```
jknack: bom file is auto-generated at release time, no need to update it manually :)

paul-hammant: Grin - then we shouldn't be checking it in :-P

PR #1008, opened 2018-02-07
upgrade aws

paul-hammant:
all tests pass


paul-hammant:
Amazon have already moved some minor versions further on. One day, MavenCentral is going to crash and burn under the weight...

PR #1011, opened 2018-02-10
upgrade jetty take 2

paul-hammant:


jknack:
This wasn't easy. See #1018

PR #1023, opened 2018-02-14
hikariCP.version upgrade

paul-hammant:


jknack:
did you run: `./build/coverage.sh`?

JooqFeature.java is failing.

paul-hammant:
I just do a 'mvn clean install' from the top. I'll do coverage.sh going forwards. Including for this one.

paul-hammant:
Yup coverage report fails for that. It's inexplicable to me though :-(

Do you debug that through Intellij with a profile set.  Does it adequately setup the classpath in the same way that the Maven command line does?  I mean so that you can quickly solve towards root cause methodically.

jknack:
Closing in favor of #1031

PR #1040, opened 2018-02-24
Clarifying the @Local injection documentation & injection error

yholkamp:
While debugging my application I ran in to some `@Local` behavior that I didn't immediately expect, in that matching happens based on the parameter name rather than type (like with injection). In hindsight this makes perfect sense and I figured I'd add a note about this in the documentation to help others who run in to this.

Additionally I noticed that currently the error you receive when you define an mvc route with has a non-optional `@Local` field, i.e. `@Local String myLocal` and this value is not present, you receive the following exception/trace, which provides little debug information:
```
java.util.NoSuchElementException: No value present
	at java.util.Optional.get(Optional.java:135)
	at org.jooby.internal.mvc.RequestParam.lambda$static$16(RequestParam.java:319)
	at org.jooby.internal.mvc.RequestParam.value(RequestParam.java:368)
```
I've added a commit that replaces this exception with a `Jooby.Err` exception that includes the route and parameter in which this occurred, which should help a lot while debugging this kind of errors. I've also added test coverage for this.

The new exception then becomes:
```
org.jooby.Err: Server Error(500): Could not find required local 'myLocal', which was required on /mypath
	at org.jooby.internal.mvc.RequestParam.lambda$static$16(RequestParam.java:315)
	at org.jooby.internal.mvc.RequestParam.value(RequestParam.java:365)
```

### DIFF
--- a/doc/req.md
+++ b/doc/req.md
@@ -390,26 +390,33 @@ Local attributes (a.k.a request attributes) are bound to the current request. Th
 }
 ```
 
-In ```mvc routes``` request locals can be injected via the ```@Local``` annotation:
+In ```mvc routes``` request locals can be injected via the ```@Local``` annotation by defining a method parameter with the same name as a request local:
 
 ```java
 
   @GET
   public Result localAttr(@Local String var) {
-    ...
+    // either var will be set to a previously configured value or an error is thrown when this method is requested
   }
 
   @GET
   public Result ifLocalAttr(@Local Optional<String> var) {
-    ...
+    // var will contain an optional that may be empty if no value was previously configured
   }
 
   @GET
   public Result attributes(@Local Map<String, Object> attributes) {
-    ...
+    // attributes contains the map of local attributes
   }
 
 ```
+
+Locals that are available by default include:
+
+* `path`: the request path, i.e. `/myroute`
+* `contextPath`: application path (a.k.a context path). It is the value defined by: `application.path`.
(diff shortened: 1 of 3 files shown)


yholkamp:
Somehow my changes have triggered a build failure in the `AssetForwardingTest`:
```
Tests run: 8, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.013 sec <<< FAILURE! - in org.jooby.AssetForwardingTest
url(org.jooby.AssetForwardingTest)  Time elapsed: 0.006 sec  <<< ERROR!
java.lang.RuntimeException: java.lang.RuntimeException: javassist.NotFoundException: java.net.URL
	at org.jooby.AssetForwardingTest.url(AssetForwardingTest.java:78)
Caused by: java.lang.RuntimeException: javassist.NotFoundException: java.net.URL
	at org.jooby.AssetForwardingTest.url(AssetForwardingTest.java:78)
Caused by: javassist.NotFoundException: java.net.URL
	at org.jooby.AssetForwardingTest.url(AssetForwardingTest.java:78)
```
on this line: https://github.com/jooby-project/jooby/blob/47653e3e1b6114f9d08ff66c86645b64bd41f906/jooby/src/test/java/org/jooby/AssetForwardingTest.java#L78

I don't believe this is related to my code changes but I'm unsure how this should be resolved or why the build is passing for master branch right now.

jknack:
Very well done. Thank you.

> AssetForwardingTest

I can;t figure it out why travis randomly fails with this test

PR #1041, opened 2018-02-25
Updating the Jdbi3 dependency version

yholkamp:
Per [the Jdbi changelog](https://github.com/jdbi/jdbi/blob/ff1ee02/RELEASE_NOTES#L54-L55), Jdbi has moved to using `jdbi3-core` rather than `jdbi3` as artifact name after 3.0.0-beta2, which is the version currently in use in Jooby. 

Since there are quite a number of bugfixes that were added after this beta, in particular relating to Kotlin and nested bean mapping, I'd like to propose that we upgrade the Jdbi3 version to the latest, which is 3.0.1 on [Maven Central](http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.jdbi%22%20).

### DIFF
--- a/modules/jooby-jdbi3/pom.xml
+++ b/modules/jooby-jdbi3/pom.xml
@@ -40,7 +40,7 @@
     <!-- JDBI -->
     <dependency>
       <groupId>org.jdbi</groupId>
-      <artifactId>jdbi3</artifactId>
+      <artifactId>jdbi3-core</artifactId>
     </dependency>
 
     <!-- Test dependencies -->
--- a/pom.xml
+++ b/pom.xml
@@ -890,7 +890,7 @@
 
       <dependency>
         <groupId>org.jdbi</groupId>
-        <artifactId>jdbi3</artifactId>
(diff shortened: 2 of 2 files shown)


yholkamp:
```Tests run: 8, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.018 sec <<< FAILURE! - in org.jooby.AssetForwardingTest
url(org.jooby.AssetForwardingTest)  Time elapsed: 0.008 sec  <<< ERROR!
java.lang.RuntimeException: java.lang.RuntimeException: javassist.NotFoundException: java.net.URL
	at org.jooby.AssetForwardingTest.url(AssetForwardingTest.java:78)
Caused by: java.lang.RuntimeException: javassist.NotFoundException: java.net.URL
	at org.jooby.AssetForwardingTest.url(AssetForwardingTest.java:78)
Caused by: javassist.NotFoundException: java.net.URL
	at org.jooby.AssetForwardingTest.url(AssetForwardingTest.java:78)```
These tests are flaky again sadly.

PR #1042, opened 2018-02-25
Updating AssetMojo to honor the Maven project.baseDir

yholkamp:
Currently if you run the `AssetMojo` from another directory than the project root, asset compilation fails. Examples where this would happen are using Jooby in a Maven module of a larger project, in that case running `mvn package` on the main project would break as `package` is invoked on the Jooby submodule from the parent directory. Another example would be perhaps on a CI server compiling the app without changing directories via `mvn -f projectdir/pom.xml package`.

I'm aware that this isn't the recommended way to run Jooby but it'd be convenient to support this nonetheless as it makes it easier to integrate Jooby in to existing projects.

Since the `AssetMojo` has access to `mavenProject.basedir` which returns the proper base directory irrespective of the working directory, I've made the following changes:
* The `AssetClassLoader` now has an extended version that takes a projectBaseDir, with an overloaded variant that defaults to the current directory.
* The `AssetCompiler` now assumes the provided classLoader is an instance that was optionally pre wrapped by `AssetClassLoader`, allowing the `AssetMojo` to pass in an `AssetClassLoader` that uses the Maven basedir.

### DIFF
--- a/modules/jooby-assets/src/main/java/org/jooby/assets/AssetClassLoader.java
+++ b/modules/jooby-assets/src/main/java/org/jooby/assets/AssetClassLoader.java
@@ -220,7 +220,7 @@ public class AssetClassLoader {
    * @throws IOException if an exception occurred
    */
   public static ClassLoader classLoader(final ClassLoader parent) throws IOException {
-    return classLoader(parent, new File(""));
+    return classLoader(parent, new File(System.getProperty("user.dir")));
   }
   
   /**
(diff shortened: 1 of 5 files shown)


modules/jooby-assets/src/main/java/org/jooby/assets/AssetClassLoader.java
```diff
@@ -210,13 +210,32 @@
...
+  
+  /**
+   * Constructs a new AssetClassLoader that includes the 'public' dir if present in the current working dir
+   *
+   * @param parent parent classloader
+   * @return classloader that includes the public dir if present
+   * @throws IOException if an exception occurred
+   */
   public static ClassLoader classLoader(final ClassLoader parent) throws IOException {
+    return classLoader(parent, new File(""));
```
jknack: use: `new File(System.getProperty("user.dir");`

yholkamp:
Thanks @jknack, good point regarding `user.dir`. I've added that in ddd9154, the tests failed on the `AssetForwardingTest` again sadly enough.

Issue #1087, opened 2018-05-14
Add Jackson view support.


huysamen:
Add support for rendering specific views during JSON serialisation when using Jackson and the `@JsonView` annotation.

huysamen:
I was thinking to minimise the impact on core Jooby, since this is Jackson specific, I can just add a `JacksonView` class which the renderer can check for before serialisation, and then apply the view if an instance of that class was passed into the `json` call.

huysamen:
#1088

PR #1088, opened 2018-05-14
Add @JsonView support to Jackson module

huysamen:
Adds `@JsonView` support to Jackson module.


huysamen:
#1087

jknack:
Like it, but think we should do it a bit different using route attributes:

Script:
```
  get("/:id", req -> {
    return ...;
  }).attr("jsonView", Views.Public.class);
```

MVC:

```
@JsonView(Views.Public.class)
public Item get(Long id) {
  ...
}
```

This way we don't have to return `new JacksonView(Views.Public.class, data)`

To do this, we have to add a new `attr(String)` method to `Renderer.Context` so we can read it from jackson renderer.

What do you think? Can you do it?

huysamen:
Sounds good, will see if I can update the PR sometime today.

huysamen:
Although, not 100% sure I am on board with annotating the method in MVC.  This limits you to only ever returning that specific view.  The use-case for JSON views is to render different views of objects based on some condition, usually roles/permissions assigned to the principal, which often only is determined after entering the controller.  So you need a way to be able to set the view to render in the method.

jknack:
Good point. Let's keep this pull open (as it is) need to think a bit more how to do it... we can probably use/reuse the existing `Result` object

huysamen:
My initial idea was adding it to the existing `Result` object, however, since this is so Jackson specific (I'm not too familiar with GSON but I don't think they support views), I was thinking "separation of concerns" and all that.  Why pollute the core subsystem with data only used by 1 module?

jknack:
You're right. Let's keep what you did.

jknack:
Going to merge and do some small changes

Issue #1102, opened 2018-05-30
Invalid response status in metrics for deferred methods


lzvtrifork:
I noticed that the Metrics module would report all responses as status 200 for deferred methods. It appears that this happens because `InstrumentedHandler` puts the meter in a _finally_ block, and `ResponseImpl` throws a `DeferredExecution` exception as a way to handle responses. This results in the _finally_ block being executed before the response has a status set, and it just defaults to 200.

I came up with this alternative implementation for the `InstrumentedHandler` (inspired by the `RequestLogger`) that seems to work. Please consider incorporating this into Jooby code.

```java
public class InstrumentedHandler implements Route.Handler {

    @Override
    public void handle(final Request req, final Response rsp) throws Throwable {

        MetricRegistry registry = req.require(MetricRegistry.class);
        Counter counter = registry.counter("request.actives");
        Timer.Context timer = registry.timer("request").time();
        counter.inc();

        rsp.complete((ereq, ersp, x) -> {
            timer.stop();
            counter.dec();
            Meter meter = registry.meter("responses." + rsp.status().orElse(Status.OK).value());
            meter.mark();
        });
    }
}
```

jknack:
Can you send a pull request?

PR #1104, opened 2018-05-30
Metrics: collect metrics after the response is complete

lzvtrifork:
Fixes invalid response status in metrics for deferred methods as reported in #1102


jknack:
This pull is a fix for #1102

PR #1123, opened 2018-06-11
Code cleanup

feffi:
Hi, fixed some sonarqube and checkstyle issues. Just code quality cleanup, eliminationg of java 5/6/7 stuff and simplifying of methods.

more to come...


jknack:
@feffi Thanks for this, just a couple of general comments:

- Next time please try to send a smaller PR. Otherwise, it is hard to do it right :+1
- Try to not change import statement order (this is usually done automatically by IDE, I know.)
- I prefer single import `java.util.List` over wide import `java.util.*`. Even when there are multiple classes from same package.

feffi:
Hi @jknack, sure thing, I tried to do it module-wise, sorry for the „huge“ PR.  Thanks for the hints, I‘ll switch my IDE settings ;-)

Does Jooby or you have any coding guidelines like the ones you mentioned on some wiki page? Would really help :-)

But a question: Why do not change statement order? Did I miss something?

jknack:
> Does Jooby or you have any coding guidelines like the ones you mentioned on some wiki page? Would really help :-)

I suppose to add checkstyle, but never did :(. Probably good time to add it? At least to the core and only on public classes? 


I mean these kind of changes:

```
package org.jooby;
 
-import static java.util.Objects.requireNonNull;
-
 import javax.annotation.Nonnull;
 import java.util.function.Supplier;
 
+import static java.util.Objects.requireNonNull;
+
```

They do nothing.

jknack:
@feffi can you please rebase? It can't be merge bc I pushed some commits (sorry)

modules/jooby-assets/src/main/java/org/jooby/assets/Assets.java
```diff
@@ -516,9 +516,8 @@ public void configure(final Env env, final Config config, final Binder binder) t
     // live compiler?
     AssetHandler handler;
     if (watch) {
-      Path publicDir = Paths.get(conf.getString("user.dir"), "public");
       Path workdir = Paths.get(conf.getString("assets.outputDir"));
-      handler = new FileSystemAssetHandler("/", workdir, publicDir);
+      handler = new FileSystemAssetHandler("/", workdir);
```
jknack: you need to revert this change. We do need multi sources for file system asset handler

modules/jooby-assets/src/main/java/org/jooby/internal/assets/FileSystemAssetHandler.java
```diff
@@ -203,30 +203,23 @@
...
 
 public class FileSystemAssetHandler extends AssetHandler {
 
-  private final List<Path> sources;
+  private final Path workdir;
 
-  public FileSystemAssetHandler(final String pattern, final Path... sources) {
+  public FileSystemAssetHandler(final String pattern, final Path workdir) {
     super(pattern);
-    this.sources = Arrays.asList(sources);
```
jknack: same here, please revert.

modules/jooby-assets/src/main/java/org/jooby/internal/assets/FileSystemAssetHandler.java
```diff
@@ -203,30 +203,23 @@
...
+    this.workdir = workdir;
   }
 
   @Override protected URL resolve(String path) throws Exception {
-    return sources.stream()
```
jknack: here two, we do resolve against multiple input dirs

modules/jooby-assets/src/main/java/org/jooby/assets/Assets.java
feffi: Hi Edgar, this change was coming from f10ac3f87 (upstream/master merge. I surely will revert it, sorry!

modules/jooby-assets/src/main/java/org/jooby/internal/assets/FileSystemAssetHandler.java
feffi: done

modules/jooby-assets/src/main/java/org/jooby/internal/assets/FileSystemAssetHandler.java
feffi: done

jknack:
Please don't change method signature on public API, here are some examples:

- Route.java: 2299
    - Object handle() throws FileNotFoundException
    + Object handle()

- Sse.java: 632/647
   - void handle(Request req, Sse sse) throws Exception;   

UnitOfWork.java:
-  default void accept(final Throwing.Consumer<Session> callback) throws Throwable {
+  default void accept(final Throwing.Consumer<Session> callback) {

You are free to remove the throws clause on implementors but please don't remove from API.

Also seems you added [xerces](https://github.com/jooby-project/jooby/commit/0033e82)?  why we need it? Also, you didn't specify a version for `xercesImpl`

I like what you did and want to merge it, but we need to make sure everything keep working as expected.

Travis build isn't working (still need to figure it out why), but can you please run `./build/coverage.sh` and confirm everything works?

Thanks

feffi:
Hi, regarding Xerces: we don't use it directly, but transitive. And it is vulnerable (as currently all versions of it). I'm trying to migrate some dependencies to not vulnerable versions. So here you go:

✗ High severity vulnerability found on xerces:xercesImpl@2.8.1
- desc: Denial of Service (DoS)
- info: https://snyk.io/vuln/SNYK-JAVA-XERCES-31497
- from: org.jooby:jooby-project@1.4.1-SNAPSHOT > org.jooby:coverage-report@1.4.1-SNAPSHOT > org.pac4j:pac4j-openid@2.2.1 > org.openid4java:openid4java@1.0.0 > xerces:xercesImpl@2.8.1

feffi:
The rest I'm rolling back. I'm getting familiar with your caller structure, so forgive me those type of "optimizations". I was just seeing your point on the second sight :-)

feffi:
for vulnerable dependencies, also see https://github.com/jooby-project/jooby/issues/1129

feffi:
Hi @jknack,

please don't merge this, I#ll rework this into smaller PRs for each mvn-project for better reviewability.

PR #1130, opened 2018-06-18
Adding support for modules with @Provides annotated methods

yholkamp:
When using dependency injection for modules that rely heavily on providers, Guice's `@Provides` support (https://github.com/google/guice/wiki/ProvidesMethods) can be used to reduce the boilerplate that's otherwise required when registering `Provider` classes. Currently these annotations are not supported by Jooby.

With this pull request I'm proposing to add a method call in the Jooby `install` method that binds these annotated methods by using existing Guice functionality to generate a module for this. If a module has no such methods, the call will essentially be a no-op.

I'm curious to hear what you think about this proposal @jknack and thanks in advance for considering it.

### DIFF
--- a/jooby/src/main/java/org/jooby/Jooby.java
+++ b/jooby/src/main/java/org/jooby/Jooby.java
@@ -220,6 +220,7 @@ import com.google.inject.Key;
 import com.google.inject.Provider;
 import com.google.inject.Stage;
 import com.google.inject.TypeLiteral;
+import com.google.inject.internal.ProviderMethodsModule;
 import com.google.inject.multibindings.Multibinder;
 import com.google.inject.name.Named;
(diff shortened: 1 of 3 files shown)


jknack:
Looks good! Did you run `mvn clean package`? 99% sure that JoobyTest is failing, but if not ignore my comment.

Thank you

yholkamp:
@jknack you're right, I skipped an essential step there. I believe commit b9c462170 patches that but TravisCI doesn't seem to have pulled in the last commit when running the tests again, so it's not showing up. Is that something you can re-run or should I recreate my PR instead?

jknack:
Looks good. Travis build (linux/ubuntu) is broken need to see why.

Going to merge later today.

PR #1150, opened 2018-06-25
code cleanup jooby-akka

feffi:
Just some code cleanup.

### DIFF
--- a/modules/jooby-akka/src/main/java/org/jooby/akka/Akka.java
+++ b/modules/jooby-akka/src/main/java/org/jooby/akka/Akka.java
@@ -215,6 +215,8 @@ import com.typesafe.config.ConfigFactory;
 
 import akka.actor.ActorSystem;
 
+import javax.annotation.Nonnull;
+
 /**
  * <h1>akka</h1>
  * <p>
@@ -265,7 +267,7 @@ public class Akka implements Module {
   }
 
   @Override
-  public void configure(@NotNull final Env env, @NotNull final Config conf, @NotNull final Binder binder) {
+  public void configure(@Nonnull final Env env, @Nonnull final Config conf, @Nonnull final Binder binder) {
     ActorSystem sys = ActorSystem.create(name, conf);
 
     env.serviceKey().generate(ActorSystem.class, name, syskey -> binder.bind(syskey).toInstance(sys));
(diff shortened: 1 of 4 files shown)


modules/jooby-akka/src/main/java/org/jooby/akka/Akka.java
```diff
@@ -205,6 +205,7 @@
 
 import static java.util.Objects.requireNonNull;
 
+import com.sun.istack.internal.NotNull;
```
jknack: use `javax.annotaion.Nonnull` and avoid use of internal classes

feffi: Hi, it seems the JSR-305 import in the POM was missing. I added it globally in the parent POM.

Any intention from your side regarding such global POM imports? A lot of imports could be moved from modules to parent POM (e.g. PowerMock, etc.). Any thoughts on that?

feffi: IntelliJ did a Fallback on NotNull in a strange way...fixed it

jknack: We shouldn't add global dependencies to parent pom.xml

In core, this was handled it like:

```
<dependency>
      <groupId>com.google.code.findbugs</groupId>
      <artifactId>jsr305</artifactId>
      <scope>provided</scope>
    </dependency>
```

feffi: This does work, if you add Java8EE manually to your IDE (mine hasn't this libs). Maybe I have a configiguration problem here? Any hints would be appreciated?

pom.xml
```diff
@@ -1681,6 +1681,14 @@
     </dependencies>
   </dependencyManagement>
 
+  <dependencies>
```
jknack: please move this to the jooby-akka/pom.xml

jknack:
fyi: travis build is back

feffi:
Oh nice, how come?

jknack:
I couldn';t figure it the easymock/javassist error while mocking URL objects. Just create a real URL object and use it. https://github.com/jooby-project/jooby/commit/25991d52e67807a1799c9f6a6bbc0be735acce30

pom.xml
feffi: done.

feffi:
I really like the green travis stuff :-)

PR #1151, opened 2018-06-25
added ReDos warning

feffi:


jknack:
How to fix it?

Still, need to do a closer look, but I believe this is a closed context which hard to exploits

feffi:
You‘re right, I just stumbled across it and wanted to market it. A colleague of mine is actually looking into it. Just keep this on hold.

PR #1171, opened 2018-06-25
cleanup-jooby-rxjava

feffi:
fixed minor code issues

### DIFF
--- a/modules/jooby-rxjava/src/main/java/org/jooby/rx/Rx.java
+++ b/modules/jooby-rxjava/src/main/java/org/jooby/rx/Rx.java
@@ -357,7 +357,7 @@ public class Rx extends Exec {
 
     private Deferred deferred;
 
-    private AtomicBoolean done = new AtomicBoolean(false);
+    private final AtomicBoolean done = new AtomicBoolean(false);
 
     public DeferredSubscriber(final Deferred deferred) {
       this.deferred = deferred;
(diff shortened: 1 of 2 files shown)


modules/jooby-rxjava/src/main/java/org/jooby/rx/Rx.java
```diff
@@ -366,7 +366,7 @@ public DeferredSubscriber(final Deferred deferred) {
     @Override
     public void onCompleted() {
       if (done.compareAndSet(false, true)) {
-        deferred.resolve((Object) null);
+        deferred.resolve(null);
```
jknack: Are you sure about this? Feel we need the cast to `Object`

jknack: You're right about `resolve(Object)`, please ignore my comment.

Are you on linux? I did see that error in Travis, still can't reproduce/fix it

feffi: no problem :-) Happy to help!

Nope, I'm on macOS (so kinda posix yeah). I found this regarding the error:

https://stackoverflow.com/questions/12190377/getting-javassist-notfoundexception-with-powermock-and-powerrule-in-junit-with-m

maybe we just need to switch to the agent version? I'm not that deep into your testing (yet), but maybe this helps?

feffi: btw: locally I can't reproduce this error neither. All tests (maven/console and IntelliJ) are fine.

PR #1185, opened 2018-06-29
Update jQuery to 2.2.4, jquery-ui to 1.12.1

DerDackel:
Hey there,
I did a little busywork upgrading jQuery to the latest stable 2.x release in the same vein as @feffi's version update crusade.
It's mostly a find & replace job, as the specific version of the artifact was referenced across a lot of different tests (maybe we could reduce that?). The whole thing is largely only relevant for testing but it posed a good starter, as I'm currently familiarizing myself with Jooby's test suites

### DIFF
--- a/modules/coverage-report/src/test/java/org/jooby/AssetFeature.java
+++ b/modules/coverage-report/src/test/java/org/jooby/AssetFeature.java
@@ -71,7 +71,7 @@ public class AssetFeature extends ServerFeature {
         .get("/assets/jquery-2.1.4.js")
         .expect(200)
         .header("Content-Type", "application/javascript;charset=UTF-8")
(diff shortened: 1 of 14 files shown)


DerDackel:
Is there some instability with the tests on Travis? The test blowing up seems totally unrelated and I'm having a serious case of "Can't reproduce" here.

feffi:
Yep, got some DNS errors yesterday

jknack:
Merged @feffi pull request and this need to be rebased now... After a closer look I realize the jquery jar is used for testing large files requests... Feel we shouldn't care what version of jquery is as long we keep testing large files, no?

DerDackel:
Yeah, that one was more my OCD. exorcised it from the commit and rebased master.

PR #1191, opened 2018-07-02
Taint analysis

feffi:
added taint marker as first step for fixing taints/sinks.


feffi:
sorry, false PR, just ignore.

PR #1192, opened 2018-07-02
Taint analysis

feffi:
added taint marker as first step for fixing taints/sinks.


jknack:
I'm not sure what is suppose to do here :)

feffi:
Oh, it‘s just the markers for possible sanitisation of outputs. Nothing immeasurable to be done, but to remember verifying and possibly fixing later.

PR #1195, opened 2018-07-06
Fix missing flush on Netty context

### LINKED ISSUE #1194: Deferred.resolve() without result body doesn't finished the request
Hi,
Currently when using `deferred.resolve(Results.ok())` or `deferred.resolve(Results.with(Status.NO_CONTENT))` or any other means to return result without any body Jooby doesn't "finished" the request and it will left open until server/client/middle man closes on timeout

Once we change to simple `deferred.resolve(Results.ok("dummy"))` everything works as expected

DerDackel:
Fixes #1194

Seems like in that case, there was no context flush occurring, thus the response stayed in the context's buffer indefinitely.

At least I hope I parsed that right as this was my first experience debugging Netty-related code.

### DIFF
--- a/modules/coverage-report/src/test/java/org/jooby/issues/Issue1194.java
+++ b/modules/coverage-report/src/test/java/org/jooby/issues/Issue1194.java
@@ -1,0 +1,22 @@
+package org.jooby.issues;
+
+import org.jooby.Results;
+import org.jooby.test.JoobySuite;
+import org.jooby.test.ServerFeature;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertTrue;
+
+@RunWith(JoobySuite.class)
+public class Issue1194 extends ServerFeature {
+
+    {
+        get("/async", promise(deferred -> deferred.resolve(Results.ok())));
+    }
+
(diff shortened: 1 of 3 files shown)


DerDackel:
CI Test logs show 

```ERROR [2018-07-07 14:56:14,226] [main] An error occurred while starting the application:
java.net.BindException: Address already in use```

followed by 

```Time elapsed: 0.18 s[1;31m <<< FAILURE![m - in org.jooby.issues.[1mIssue555[m
[[1;31mERROR[m] rawPath[netty](org.jooby.issues.Issue555)  Time elapsed: 0.015 s  <<< ERROR!
org.apache.http.conn.HttpHostConnectException: Connect to localhost:46676 [localhost/127.0.0.1] failed: Connection refused (Connection refused)
    at org.jooby.issues.Issue555.rawPath(Issue555.java:17)
Caused by: java.net.ConnectException: Connection refused (Connection refused)
    at org.jooby.issues.Issue555.rawPath(Issue555.java:17)```

Could this hark back to a timing problem with the preceding test or are we still leaking something?

jknack:
Going to check soon, but yea Travis is hard sometimes.

PR #1202, opened 2018-07-09
Sonarqube cleanup

feffi:
some code simplification
removed deprecated Throwables.propagate

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,49 @@
 language: java
 
-before_install:
-  - export MAVEN_OPTS=-Xmx1024m
-
-script:
-  - ./build/travis.sh
-
 jdk:
   - oraclejdk8
+  - oraclejdk9
+  - oraclejdk10
+  - openjdk7
+  - openjdk8
+  - openjdk9
+  - openjdk10
+  - openjdk11
+
+stages:
+  - package
+  - test
+  - coverage
+
+cache:
+  directories:
+  - $HOME/.m2
+
+#before_install:
+#  - export MAVEN_OPTS=-Xmx1024m
+
+jobs:
+  include:
+    - stage: package
+      env:
+        - MAVEN_OPTS=-Xmx1024m -DskipTests=true -Dlogback.configurationFile=logback-travis.xml
+      script:
+        - "mvn clean install"
+#    - stage: test
+#      env:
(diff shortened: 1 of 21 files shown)


.travis.yml
```diff
@@ -1,13 +1,49 @@
...
 
-before_install:
-  - export MAVEN_OPTS=-Xmx1024m
-
-script:
-  - ./build/travis.sh
-
 jdk:
   - oraclejdk8
+  - oraclejdk9
```
jknack: build don't work on Java9 or higher

jooby/src/main/java/org/jooby/Asset.java
```diff
@@ -324,7 +326,13 @@ default String etag() {
     b.append("W/\"");
 
     BaseEncoding b64 = BaseEncoding.base64();
-    int lhash = resource().hashCode();
+    int lhash = 0;
+    try {
+      lhash = resource().toURI().hashCode();
```
jknack: why this change?

jooby/src/main/java/org/jooby/Asset.java
```diff
@@ -324,7 +326,13 @@ default String etag() {
...
+    } catch (URISyntaxException e) {
+      throwIfUnchecked(e);
+      throw new RuntimeException(e);
```
jknack: URISyntaxException isn't an unchecked exception and I don't like to wrap exception as runtime. It is just noise

jooby/src/main/java/org/jooby/Err.java
```diff
@@ -462,17 +462,17 @@ public int statusCode() {
    * @return A lightweight view of the err.
    */
   public Map<String, Object> toMap(boolean stacktrace) {
-    Status status = Status.valueOf(this.status);
+    Status statusOutput = Status.valueOf(this.status);
```
jknack: why the name change? This is when sonar or similar tools don't make sense.

.travis.yml
feffi: Oh sorry, this change wasn't meant to be in this PR. Just disable it.

feffi:
obsolete

jknack:
Yea, I applied some of your changes manually to core (mainly).

Thanks

PR #1211, opened 2018-07-14
Fix broken module name badges

yasuyk:
A backslash needs to be escaped by a backslash in status.

see https://shields.io/#your-badge


yasuyk:
@jknack Thank you for merging! 😄

PR #1250, opened 2018-10-16
Fix documentation

gcusnieux:


jknack:
Can you make the change [here](https://github.com/jooby-project/jooby/blob/master/doc/doc/assets-autoprefixer/autoprefixer.md)?

Thanks

Issue #1197, opened 2018-07-07
add sonarcloud analysis


feffi:
@see https://sonarcloud.io/

jknack:
found impossible to push from Travis. I tried all without lock keep getting:

> You're only authorized to execute a local (preview) SonarQube analysis without pushing the results to the SonarQube server. Please contact your SonarQube administrator. -> [Help 1]
[ERROR] 

I did pushed from my laptop (without coverage), but isn't the same. Travis should works too

https://sonarcloud.io/dashboard?id=org.jooby%3Ajooby-project%3Amaster

feffi:
I'll look into this.

JeffLeFoll:
Hello
I had a look to this issue, I managed to have travis pushed data to sonarcloud.
You can have a look to my jooby's fork results here : https://sonarcloud.io/dashboard?id=jooby-sonarcloud
The only remaining issue I have is that the coverage report are not at all the same as the coveralls one.

JeffLeFoll:
Here is the Sonar report from travis : https://sonarcloud.io/dashboard?id=jfl-sonar-jooby-test

marcelstoer:
I guess you know that but you may use `<sonar.exclusions>` and `<sonar.coverage.exclusions>` for that.

https://docs.sonarqube.org/latest/project-administration/narrowing-the-focus/

jknack:
Outdated

PR #1253, opened 2018-10-21
add sonarcloud analysis from travis

JeffLeFoll:
Hi, this PR is related to issue #1197 .

You can see the resulting sonar analysis result here : https://sonarcloud.io/dashboard?id=jfl-sonar-jooby-test

Please not that in order to work you need to set in Travis an Env. variable called SONAR_TOKEN with your token configured in sonarcloud.

The code coverage % is not the same in sonar and coveralls because coveralls is configure to only analyse some packages and sonar take the whole sources in account.

It maybe possible to configure sonar to only take a sub-set of the sources in account.


Daniel-Dos:
very good .

Daniel-Dos:
@jknack please review .

jknack:
Thank you, for this. I tried a lot without luck, I was missing the `sonar.jacoco.reportPaths` variable.

JeffLeFoll:
You're welcome, I'm glad to have been helpful :)

But just to make it clear, with this configuration you will have a huge difference on the % of coverage between sonar and coveralls.

This is due to the fact that coveralls is configured to only take into account a small set of jooby's source whereas sonar will take into account all the sources (for instance event modules without tests).

It may be possible to configure sonar to excludes some sources.

jknack:
Make senses, need to work a bit more to sync both coverage. The code we skip is basically:

- code I took from somewhere else, like Netty SSL
- awful code from ApiTool (bytecode parsing)

JeffLeFoll:
Ok, I will try to add the configuration to exclude some sources that don't make sense.

jknack:
Thanks. These are the classes we skip: https://github.com/jooby-project/jooby/blob/master/modules/coverage-report/pom.xml#L202-L213

As said, not all the project are covered. Complete list of covered projects is here: https://github.com/jooby-project/jooby/blob/master/modules/coverage-report/pom.xml#L39-L94

PR #1256, opened 2018-10-23
update dependencies

Daniel-Dos:
update some dependencies.


jknack:
Thanks for this, looks good. But now we are two here, let's wait for feedback before merging.

Thanks again.

Daniel-Dos:
great .  :  )

Issue #1281, opened 2019-01-23
Redis Sentinel support


AmaranthLIS:
Redis can work in [cluster mode](https://redis.io/topics/sentinel) with Sentinel and Jedis client support that, but current integration jooby-jedis doesn't implement that part.

This is really important thing if you need high availability with Redis.

AmaranthLIS:
I've send pull request.
Honestly what kind of information about License need to put, i've copied yours. 
I hope it's okey.
May some changes need to do, just give me know.

jknack:
Done in #1282

PR #1282, opened 2019-01-23
Redis Sentinel support

AmaranthLIS:
Pull request for issue #1281

### DIFF
--- a/modules/jooby-jedis/src/main/resources/org/jooby/jedis/jedis.conf
+++ b/modules/jooby-jedis/src/main/resources/org/jooby/jedis/jedis.conf
@@ -24,3 +24,7 @@ jedis.pool.jmxNamePrefix = redis-pool
 # session store, key prefix and timeout in seconds
 jedis.session.prefix = sessions
 jedis.session.timeout = ${session.timeout}
+
+jedis.sentinel.hosts = ["localhost:26379"]
+jedis.sentinel.master = "master"
+jedis.password = ""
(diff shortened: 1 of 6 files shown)


modules/jooby-jedis/src/main/resources/org/jooby/jedis/jedis.conf
```diff
@@ -24,3 +24,7 @@ jedis.pool.jmxNamePrefix = redis-pool
 # session store, key prefix and timeout in seconds
 jedis.session.prefix = sessions
 jedis.session.timeout = ${session.timeout}
+
+jedis.sentinel.hosts = ["localhost:26379"]
```
jknack: don't think we should have default values for all these variables

jknack:
hey, looks very good, some minor docs setup is missing but I can do that.

Thank you

modules/jooby-jedis/src/main/resources/org/jooby/jedis/jedis.conf
AmaranthLIS: For Redis Sentinel doesn't use these settings

> jedis.session.prefix = sessions
> jedis.session.timeout = ${session.timeout}

About this setting
> jedis.sentinel.hosts = ["localhost:26379"]

We can make it empty and check in constructor.

jknack: yea, just want to avoid a default connection or connection where developer didn't notice. In order to connect they must provide the correct settings

AmaranthLIS: I've did it.

PR #1304, opened 2019-03-21
Upgrade to Hibernate 5.3.9

imeszaros:
This pull request upgrades `jooby-hbm` to use the latest stable version of Hibernate in the 5.3 series.

Beside simply bumping the version number, two additional changes were necessary:

- they marked the CDI-API is provided dependency, therefore `jooby-hbm` needs to depend on it explicitly: https://hibernate.atlassian.net/browse/HHH-11370
- the usage of `JpaIntegrator` is not necessary anymore, so it can be removed from the initialization code and tests: https://hibernate.atlassian.net/browse/HHH-11264

### DIFF
--- a/modules/jooby-hbm/pom.xml
+++ b/modules/jooby-hbm/pom.xml
@@ -37,10 +37,16 @@
       <version>${project.version}</version>
     </dependency>
 
+    <!-- CDI Api; https://hibernate.atlassian.net/browse/HHH-11370 -->
+    <dependency>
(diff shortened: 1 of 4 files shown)


imeszaros:
A third change was also necessary:

- `BeanManager` has to be registered at bootstrapping time: https://docs.jboss.org/hibernate/orm/5.3/javadocs/org/hibernate/cfg/AvailableSettings.html#CDI_BEAN_MANAGER

PR #1368, opened 2019-08-14
Avoiding possible XSS attack through the default error handler

imeszaros:
@jknack  Please review, especially the test case. I'm not familiar with it, i'm not sure if this line is correct:

`expect(env.xss("html")).andReturn(HtmlEscapers.htmlEscaper()::escape);`

### DIFF
--- a/jooby/src/main/java/org/jooby/Err.java
+++ b/jooby/src/main/java/org/jooby/Err.java
@@ -203,10 +203,7 @@
  */
 package org.jooby;
 
-import java.util.LinkedHashMap;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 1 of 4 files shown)


jooby/src/main/java/org/jooby/Err.java
```diff
@@ -287,12 +291,24 @@ public void handle(final Request req, final Response rsp, final Err ex) throws T
...
       Config conf = req.require(Config.class);
+      Env env = req.require(Env.class);
       boolean stackstrace = Try.apply(() -> conf.getBoolean("err.stacktrace"))
-          .orElse(req.require(Env.class).name().equals("dev"));
+          .orElse(env.name().equals("dev"));
+
+      Function<Object, String> xssFilter = env.xss("html").compose(Objects::toString);
+      BiFunction<String, Object, String> escaper = (k, v) -> xssFilter.apply(v);
+
+      Supplier<Map<String, Object>> detailsProvider = () -> {
```
jknack: let's remove the supplier and just use the Map

jknack: let's remove the supplier and just use the Map

jknack: let's remove the supplier and just use the Map

imeszaros: One question about this: in the original code, the `html` and `all` cases got two different instances of the map, because `toMap()` was invoked in the `Supplier` lambda passed to the `when` calls:

```
Results
    .when(MediaType.html, () -> Results.html(VIEW).put("err", ex.toMap(stackstrace)))
    .when(MediaType.all, () -> ex.toMap(stackstrace)));
```

This is the reason I introduced the supplier. If I pass the same map instance to the suppliers in `when` calls, test cases will fail because `org.jooby.DefaultErrHandlerTest#checkErr` is invoked when validating the `html` response type view, and it removes all entries from the map, but later there is another assertation for the `all` response data which expects the size of the map to be 4.

I could either modify the test case or pass a copy of the map to the suppliers like this:

```
      Map<String, Object> details = ex.toMap(stackstrace);
      details.compute("message", escaper);
      details.compute("reason", escaper);

      rsp.send(
          Results
              .when(MediaType.html, () -> Results.html(VIEW).put("err", new LinkedHashMap<>(details)))
              .when(MediaType.all, () -> new LinkedHashMap<>(details)));
```

jknack: Got it. If you feel strongly go and modify the existing test, otherwise let's keep it

Thanks

imeszaros:
I modified the test case very slightly so `org.jooby.DefaultErrHandlerTest#checkErr` now works with a local copy of the map.

PR #1372, opened 2019-08-20
Added support for Pebble template engine

sojin:
Added support for Pebble template engine

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -260,6 +260,12 @@
 
       <dependency>
         <groupId>io.jooby</groupId>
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -403,6 +404,13 @@
         <version>${handlebars.version}</version>
       </dependency>
 
+      <!-- Pebble -->
```
jknack: add jooby-pebble into the dependency management section

sojin:
Required changes commited.

PR #1373, opened 2019-08-21
For MVC Route, be able to keep custom annotation (used for business logic)

JeffLeFoll:
Hello, 

Maybe I have miss something but today, if we put some annotation on an MVC Route, it is "lost" at runtime.

For instance if you put an annotation Protected on a route, then in a decorator you can't retrieve it to check if the call can be made (using JWT or Macaroon).

This PR give access, at Route level, to custom annotation set on MVC Route.
This PR have no effect on Script API Route.


jknack:
Thanks for the pull. This is already supported but in using an unified API for script and mvc. It is (more or less) documented here: https://jooby.io/v1/doc/#routes-properties

```java
@Target({ElementType.METHOD, ElementType.TYPE, ElementType.ANNOTATION_TYPE })
@Retention(RetentionPolicy.RUNTIME)
public static @interface Role {
  String value();
}

@Path("/path")
public class AdminResource {

  @Role("admin")
  public Object doSomething() {
    ...
  }

}

{
  use("*", (req, rsp) -> {
    System.out.println(req.route().attributes())
  });
}
```

Route.attributes have any annotation added to the mvc method a simple `Map<String, String>`

JeffLeFoll:
Hello, thank you for your answer.
I just had a quick look but it doesn't seem to work (with the 2.0.6-SNAPSHOT).

I found the `ctx.getAttributes()` but in my quick test it was always null and I do not have a `getAttributes()` when I look at `ctx.getRoute()`.

I will take the time to have a deeper look tomorrow morning.

jknack:
Apologies, I missed the 2.x branch, was thinking this was for 1.x.

You're right there isn't anything yet for 2.x but we should follow something similar to what we have in 1.x. So script/mvc routes shared the same API

JeffLeFoll:
You are right, I understand.
I will look tomorrow how I can change what I have done to offer the same API.

JeffLeFoll:
Hello, 

So I have reworked my code to provide the same API for MVC and Script API.  
I also change my terminology to use the same one as in Jooby 1.x.
```
{
  get("/foo", ctx -> "Foo")
    .attribute("foo", "bar");

  decorator(next -> ctx -> {
    User user = ...;
    String role = ctx.getRoute().attribute("Role"); 
    
    if (user.hasRole(role)) {
        return next.apply(ctx);
    }      
    
    throw new StatusCodeException(StatusCode.FORBIDDEN);                              
  });
}
```
An I have updated the Router documentation to reflect the changes.

(And I may have messed up my commit history when syncing my fork...)

jknack:
your work looks good, but you will need to fix your pull and remove all the sync commits.
We only need: https://github.com/jooby-project/jooby/pull/1373/commits/8b0f78ecf03627b3ca93ea9632858b5232eb68d3

A simple test for script/mvc route will be nice too!

Thanks

JeffLeFoll:
Replace by #1374 
Sorry for the inconvenience.

PR #1374, opened 2019-08-22
Provide Route attributes for MVC and Script API.

JeffLeFoll:
Hello, 

Follow from #1373 

So I have reworked my code to provide the same API for MVC and Script API.  
I also change my terminology to use the same one as in Jooby 1.x.
```
{
  get("/foo", ctx -> "Foo")
    .attribute("foo", "bar");

  decorator(next -> ctx -> {
    User user = ...;
    String role = ctx.getRoute().attribute("Role"); 
    
    if (user.hasRole(role)) {
        return next.apply(ctx);
    }      
    
    throw new StatusCodeException(StatusCode.FORBIDDEN);                              
  });
}
```
I have updated the Router documentation to reflect the changes.
And I have added some tests / examples.

### DIFF
--- a/jooby/src/main/java/io/jooby/internal/mvc/MvcAnnotationParserBase.java
+++ b/jooby/src/main/java/io/jooby/internal/mvc/MvcAnnotationParserBase.java
@@ -5,10 +5,16 @@
  */
 package io.jooby.internal.mvc;
 
+import io.jooby.SneakyThrows;
+
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public abstract class MvcAnnotationParserBase implements MvcAnnotationParser {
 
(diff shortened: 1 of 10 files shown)


jooby/src/main/java/io/jooby/internal/mvc/MvcAnnotationParserBase.java
```diff
@@ -50,4 +56,45 @@
     }
     return result;
   }
+
+  protected Map<String, Object> attributes(Method method) {
+    Map<String, Object> attributes = new LinkedHashMap<>();
+
+    List<Annotation> specificAttributes = Stream.of(method.getDeclaredAnnotations())
+        .filter(annotation ->
+            !"io.jooby.annotations".equals(annotation.annotationType().getPackage().getName())
```
jknack: 👍

jknack:
love it, very nice work

JeffLeFoll:
Thank you.
It's a very little "give back", we really like to work with Jooby (1.x and now 2.x 😉 )

jknack:
hehe, hope to see more from you :)

What do you think if we make the `attributes` map case-insensitive or apply the same case type to script and mvc (this is what I did in 1.x)

For example:

```java
  get()
    .attribute("foo", "bar");
```
Isn't the same as:

```
@Foo("bar")
public String method() {
  ...
}
```

JeffLeFoll:
That's right, currently it's not the same.

I have no pros or cons to make the key case-insensitive, but if we do so I'm wondering if we still need to have `ctx.getRoute().getAttributes` which return the full map.
Because in this case, users (developer) will need to know that the keys where normalized upon insertion into the map.

But on the other hand, if I have a mix of MVC and Script routes, I should not forget to be aware of the key case.
Or write something like :
 ```
get()
    .attribute(Foo.class.getSimpleName(), "bar");

ctx.getRoute().attribute(Foo.class.getSimpleName());
```

jknack:
Going to merge and make a decision about case key

PR #1389, opened 2019-09-16
[v1] Test fixes for windows path separators, line endings

altmind:
This fixes:

1. Failure to start Jooby in JoobyRunner should fail the suite early - otherwise the tests will fail later with cryptic exception `org.apache.http.conn.HttpHostConnectException: Connect to localhost:9999 [localhost/127.0.0.1, localhost/0:0:0:0:0:0:0:1] failed: Connection refused: connect`

2. Assets pipeline tests fail if there is no `assets.conf` in test path. Empty assets.conf seems to fix this.

3. Relax json tests to accept both "\r\n" and "\n" in output for windows(jackson seems to return platform-dependent line endings)

4. In some places(css sprites) we should not use system path separator - the css sprites results are URLS, the path separator there is always /.
In some places (tmpdir), we should use system path separator so our tests dont fail

Before the change: 9 errors, 19 failures https://altio.us/files/Test%20Results%20-%20TestAllorg_jooby.html

After the change:  https://altio.us/files/Test%20Results%20-%20TestAllorg_jooby2.html

(I dont know how to fix whoops, sorry)

### DIFF
--- a/modules/coverage-report/src/test/java/org/jooby/jackson/JsonCustomMapperFeature.java
+++ b/modules/coverage-report/src/test/java/org/jooby/jackson/JsonCustomMapperFeature.java
@@ -1,5 +1,6 @@
 package org.jooby.jackson;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertSame;
 
 import java.net.URISyntaxException;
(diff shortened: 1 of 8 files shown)


modules/coverage-report/src/test/java/org/jooby/jackson/JsonCustomMapperFeature.java
```diff
@@ -31,12 +32,12 @@
...
   @Test
   public void get() throws URISyntaxException, Exception {
-    request()
-        .get("/members")
-        .expect("[ {\n" +
-            "  \"id\" : 1,\n" +
-            "  \"name\" : \"pablo\"\n" +
-            "} ]");
+      request().get("/members").expect(value ->
+              assertEquals("[ {\n" +
```
jknack: the `.expect` is part of Jooby testing code... a better option is to make the change there and not per each test

PR #1393, opened 2019-10-04
Ability to use SessionBuilder with jooby-hbm

### LINKED ISSUE #1319: 2.x Nullpointer when requesting resource that does not exist
While playing around with the 2.x branch I noticed that using the `Jooby.resource(ResourceKey)` method with a key that does not exists results in a NPE. I think the framework should throw a more specific error here to notify us what's wrong.

```
java.lang.NullPointerException
	at io.jooby.Router.resource(Router.java:143)
	at io.jooby.Jooby.resource(Jooby.java:448)
	at net.test.api.App.lambda$new$baa66101$1(App.java:25)
	at io.jooby.internal.handler.SendCharSequence.apply(SendCharSequence.java:32)
	at io.jooby.internal.handler.WorkerHandler.lambda$apply$0(WorkerHandler.java:33)
	at io.netty.util.concurrent.DefaultEventExecutor.run(DefaultEventExecutor.java:66)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:906)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)
```

### LINKED ISSUE #1392: Unable to use SessionBuilder with jooby-hbm
In our project we need to wire-in a session scoped hibernate interceptor. As far as i know this is currently not possible with `jooby-hbm`, since it must be done using `SessionFactory.withOptions().<your config calls here like .interceptor(...) >.openSession()`, and jooby-hbm's `SessionProvider` and `UnitOfWorkProvider` directly calls to `SessionFactory.openSession()`.

I have a workaround as described here: https://groups.google.com/forum/#!topic/jooby-project/qG6O4zKBHGI

Can I make a PR based on it?

imeszaros:
Supposed to fix #1392 

For cleannes and better testability, I created an explicit interface for the configurer callback. I tried to adapt the test cases also.

### DIFF
--- a/modules/jooby-hbm/src/main/java/org/jooby/hbm/Hbm.java
+++ b/modules/jooby-hbm/src/main/java/org/jooby/hbm/Hbm.java
@@ -209,6 +209,7 @@ import com.google.inject.name.Names;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 import org.hibernate.Session;
+import org.hibernate.SessionBuilder;
 import org.hibernate.SessionFactory;
 import org.hibernate.boot.Metadata;
 import org.hibernate.boot.MetadataSources;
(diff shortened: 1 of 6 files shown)


imeszaros:
Is there a schedule for the next maintenance release? Soon we need to put our new development (which is relying on these changes) to production.

PR #1412, opened 2019-11-11
Remove unused dependency

cesarsotovalero:
Hello. I noticed that dependency `jakarta.ws.rs-api` is declared but not used in module `jooby`. Therefore, it makes sense to remove this dependency.

### DIFF
--- a/jooby/pom.xml
+++ b/jooby/pom.xml
@@ -183,13 +183,6 @@
       <artifactId>config</artifactId>
     </dependency>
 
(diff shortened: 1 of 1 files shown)


jknack:
It was use it when we did mvc base using reflection... but think you are right now and we don't need it anymore (we do bytecode generation for mvc)

cesarsotovalero:
So, it is ready to merge, right?

Issue #1413, opened 2019-11-11
CORS Handler always throws 403 forbidden, when PUT or DELETE is used as HTTP Verb.


jerouris:
Found a possible back in CORS. When PUT or DELETE is used as HTTP Verb, cors will always throw as 403 forbidden. Here: 
isPreflight() will return false because it is not OPTIONS
(https://github.com/jooby-project/jooby/blob/5bd3b293802651b84915cb3d14bdb71f0d5a23b8/jooby/src/main/java/io/jooby/CorsHandler.java#L112-L114)
and isSimple() will return false (not POST, GET or HEAD)
(https://github.com/jooby-project/jooby/blob/5bd3b293802651b84915cb3d14bdb71f0d5a23b8/jooby/src/main/java/io/jooby/CorsHandler.java#L83-L87)
thus return 403, forbidden:
(https://github.com/jooby-project/jooby/blob/5bd3b293802651b84915cb3d14bdb71f0d5a23b8/jooby/src/main/java/io/jooby/CorsHandler.java#L62-L81)

jknack:
my understanding is that either you do a simple cors request (GET, POST, HEAD) or you need a preflight

jknack:
Look at simple Requests: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

jerouris:
According to Chrome network monitor, here is the preFlight (verb OPTIONS):

OPTIONS /api/v1/machines/123 HTTP/1.1
Host: localhost:8080
Connection: keep-alive
Access-Control-Request-Method: PUT
Origin: http://localhost:4200
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.97 Safari/537.36
DNT: 1
Access-Control-Request-Headers: authorization
Accept: */*
Sec-Fetch-Site: same-site
Sec-Fetch-Mode: cors
Referer: http://localhost:4200/machines
Accept-Encoding: gzip, deflate, br
Accept-Language: el-GR,el;q=0.9,en;q=0.8

Which returns 200 OK (Correct)

and then the second (actual) request:

PUT /api/v1/machines/123 HTTP/1.1
Host: localhost:8080
Connection: keep-alive
Content-Length: 0
Accept: application/json, text/plain, */*
Origin: http://localhost:4200
[...]

PR #1414, opened 2019-11-11
Bugfix/issue1413

jerouris:
Tries to fix issue #1413

### DIFF
--- a/jooby/src/main/java/io/jooby/CorsHandler.java
+++ b/jooby/src/main/java/io/jooby/CorsHandler.java
@@ -11,6 +11,8 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Handle preflight and simple CORS requests. CORS options are set via: {@link Cors}.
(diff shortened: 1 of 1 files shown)


jknack:
Cors tests are failing

jerouris:
Test appears to be wrong. According to MDN, OPTIONS without the access-control-request-method header, is not considered as pre-flight: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS#Examples

However, test expects this request to be treated as pre-flight:
https://github.com/jooby-project/jooby/blob/5bd3b293802651b84915cb3d14bdb71f0d5a23b8/tests/src/test/java/io/jooby/FeaturedTest.java#L2719-L2730

Use of verb OPTIONS, without the access-control-request-method header should be treated by Jooby in a special manner, providing the communication options for the target resource. 

eg: (From MDN)

To find out which request methods a server supports, one can use curl and issue an OPTIONS request:

`curl -X OPTIONS http://example.org -i`

The response then contains an Allow header with the allowed methods:

HTTP/1.1 204 No Content
Allow: OPTIONS, GET, HEAD, POST
Cache-Control: max-age=604800
Date: Thu, 13 Oct 2016 11:45:00 GMT
Expires: Thu, 20 Oct 2016 11:45:00 GMT
Server: EOS (lax004/2813)
x-ec-custom-error: 1

PR #1416, opened 2019-11-12
Updated tests in order to properly execute in windows enviroments

jerouris:
Fixing Path Separators and tackle git ability to checkout windows format (/n/r) and checkin linux (/n)


jknack:
fyi https://github.com/jooby-project/jooby/actions

Issue #1443, opened 2019-11-24
Feature request: SessionRequest Decorator


Blquinn:
I'm using the hibernate module currently and was noticing that there is a TransactionRequest Decorator which wraps each request in a transaction. However for my use case, I really just want each request to have access to the same session, not necessarily a transaction because I prefer to have more control over my transactions. Therefore I wrote a SessionRequest decorator which does just that. I modeled it after the Tx one that lives in jooby today.

```kotlin
class SessionRequest(
    private val key: ServiceKey<SessionFactory> = ServiceKey.key(SessionFactory::class.java)
): Decorator {
    override fun apply(next: Route.Handler): Route.Handler {
        return Route.Handler { ctx: Context ->
            val sessionFactory = ctx.require(key)

            var session: Session? = null

            try {
                session = sessionFactory.openSession()
                ManagedSessionContext.bind(session)
                next.apply(ctx)
            } finally {
                session?.close()
            }
        }
    }
}
```

I was wondering if there are any inherent problems with this design. If not, is there any desire to add this to jooby itself? I would imagine that others would like to have something like this.

I've also created a TransactionalMixin which allows you to wrap a block of code with a simple lambda, committing the transaction if there was no exception thrown in the block.

```kotlin
abstract class TransactionalMixin(private val sessionFactory: SessionFactory) {
    fun <T> transactional(sess: Session = sessionFactory.currentSession, fn: (sess: Session) -> T): T {
        val trans = sess.beginTransaction()
        return try {
            val retVal = fn(sess)
            trans.commit()
            retVal
        } catch (ex: Throwable) {
            trans.rollback()
            throw ex
        }
    }
}
```

Which allows you to write something like:

```kotlin
    fun createUser(request: CreateUserRequest): UserDTO {
        return transactional {
            userRepository.save(User.fromRequest(request))
           // Do something that might throw to rollback the tx.
        }.toDTO() // access returned value from lambda.
    }
```

I'd be happy to PR either, or both of those concepts into jooby if there's a desire!

Happy Sunday.

jknack:
PR for SessionRequest (+documentation) is welcome. Still to be completely safe will add a check for active transaction before closing the session? will be nice to rollback any active transaction and log an error describing the leak/bad usage.

I'm not so sure about Transactional mixin due it is for Kotlin only.

Blquinn:
Ok, I re-wrote SessionRequest in java. Will look into documentation later. 

[...]

PR #1449, opened 2019-11-26
Add SessionRequest decorator in hibernate module

Blquinn:
Note that the TransactionRequest now uses SessionRequest to add the session to the managed context.

ISSUE https://github.com/jooby-project/jooby/issues/1443

### DIFF
--- a/modules/jooby-hibernate/src/main/java/io/jooby/hibernate/SessionRequest.java
+++ b/modules/jooby-hibernate/src/main/java/io/jooby/hibernate/SessionRequest.java
@@ -1,0 +1,103 @@
+/**
+ * Jooby https://jooby.io
+ * Apache License Version 2.0 https://jooby.io/LICENSE.txt
+ * Copyright 2014 Edgar Espina
+ */
+package io.jooby.hibernate;
+
+import io.jooby.Route;
+import io.jooby.ServiceKey;
+import org.hibernate.Session;
+import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
+import org.hibernate.context.internal.ManagedSessionContext;
+import org.hibernate.resource.transaction.spi.TransactionStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Attach {@link Session} and {@link javax.persistence.EntityManager} to the current request.
+ *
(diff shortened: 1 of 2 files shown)


modules/jooby-hibernate/src/main/java/io/jooby/hibernate/TransactionalRequest.java
```diff
@@ -58,43 +56,39 @@
...
-    return ctx -> {
+    return sessionRequest.apply(ctx -> {
       SessionFactory sessionFactory = ctx.require(key);
       Transaction trx = null;
       try {
-        Session session = sessionFactory.openSession();
-        ManagedSessionContext.bind(session);
+        Session session = sessionFactory.getCurrentSession();
         trx = session.getTransaction();
-        trx.begin();
```
jknack: seems wrong, why did you remove the begin trx?

modules/jooby-hibernate/src/main/java/io/jooby/hibernate/SessionRequest.java
```diff
@@ -0,0 +1,96 @@
...
+    return (ctx) -> {
+      SessionFactory sessionFactory = ctx.require(key);
+      try {
+        Session session = sessionFactory.openSession();
+        ManagedSessionContext.bind(session);
+        Object result = next.apply(ctx);
+
+        Transaction transaction = session.getTransaction();
+        if (transaction.getStatus() == TransactionStatus.ACTIVE) {
+          transaction.rollback();
```
jknack: add log.error indicating this is not an expected state and shouldn't exist an active transaction

modules/jooby-hibernate/src/main/java/io/jooby/hibernate/TransactionalRequest.java
Blquinn: I meant to change this to session.beginTransaction(); Which creates a transaction, begins and returns that tx.

Blquinn:
I made those requested changes and amended the commit.

Blquinn:
Any idea why checkstyle is failing? I see that you committed a fix for it, but still seems to fail on this branch.

jknack:
do a build like:

```
./etc/build.sh
```

you will see the error at the console

Blquinn:
Ok, found the error. Build is running now. :)

imeszaros:
I really like this PR, it makes possible to implement a similar control flow to what we currently have in our Jooby 1.x projects using the `UnitOfWork` interface.

Issue #1552, opened 2020-03-08
Annotations inheritance in MVC routes


franklx:
Annotations on methods of base class are not preserved in children.
```kotlin
open class Base {
    @GET fun get() {}
    @POST fun post() {}
    @GET("{id}") fun get_id(@PathParam id: Int) {}
    @GET @Path("/alt/{id}") fun get_path_id(@PathParam id: Int) {}
}

@Path("/child") class Child: Base() {
    @GET("/children") fun children() {}
}
```
the above code generates:
```
  GET  /child/children
```
in jooby v1 generated:
```
  GET  /child
  POST /child
  GET  /child/{id}
  GET  /child/alt/{id}
  GET  /child/children
```
I wanted to fix it but I'm not really up on annotations processing.

franklx:
Annotations on inherited methods doesn't work because of annotation processing added in jooby-2.x (jooby-1.0 used runtime annotation evaluation).
I developed an utility class that clones inherited methods with attached annotations: https://github.com/franklx/jooby/commit/e29ab99de98e2f2280a9dcdc456f41d2b3506aa5 .
Routes are generated correctly and jooby-2.x behaviour is coherent with 1.x.
I'll write unit tests and then submit a pull request.

jknack:
Cool, will wait for it

csisy:
Sorry to comment on a closed issue, but it seems like it's not working as described in this issue. Example:

```
public class MyClass {

	@Path("concrete")
	public static class ConcreteController1 extends MyAbstractController<Whatever> {}

	public abstract static class MyAbstractController<T> {
	
		@GET
		public List<T> getAll() { ... }
	}
}
```

Te expected behaviour is to have a `GET concrete` endpoint, but instead the startup code throws an exception that 

`io.jooby.Usage: Router not found: [...]. Make sure Jooby annotation processor is configured properly.`

Which is also thrown when no route is found under a given class.

PR #1555, opened 2020-03-12
 Annotations inheritance in MVC routes #1552

franklx:
Implementation works as expected and supports inherited MVC routes as Jooby-1.x.
Commit includes passing test for issue #1552 .


franklx:
Oh, well... I'm quite stuck!
I didn't know that ```com.sun.tools.*``` isn't available in jdk8.
Any advice is welcome...

jknack:
I didn't look closely... but yes we can;'t use `com.sun` due it is an internal API

There should be a way of doing what you did using the public API: `javax.lang.model.*`

jknack:
An alternative solution will be to call all the types as we do today and before we finish check for type and super/sub type and merge and remove as accordingly.

but again I did look yet

franklx:
Public API doesn't seem to allow instantiating or cloning ```Element```s.
A possible solution is to modify ```class HandlerCompiler``` forcing the child class as "mock" owner.
I've investigated that possibility before but the Symbol API offered a cleaner solution.
I'll look at it.

franklx:
I managed removing the internal class dependencies processing classes at the end as you suggested.
It all works perfectly except of a class name problem (see travis log).
I'm not sure how to correct the problem: should I fix the test code or should I change the class name removing parent?
Assertion failing is:
```java
assertEquals(factoryName, new String(Files.readAllBytes(services), StandardCharsets.UTF_8).trim());
```
in ```jooby-apt/src/test/java/io/jooby/apt/MvcModuleCompilerRunner.java```.

Thank you

franklx:
Finally I've changed the META-INF/services file generation limiting to the first generated class.
Please let me know if this can be good.
Thank you

jknack:
Think we should generated it as long as the controller class is not abstract..hmmm... or probably better to generate leaf subclass?

If we generate for none abstract class the binding is up to final user: `mvc(new Controller())`, drawback is that we may have N generated class if the parent class isn't abstract while the leaf subclass generation generates a single entry per each leaf subclass.

franklx:
What I did is wrong and only works only with abstract classes.
META-INF/services generation must be left alone, I'll try fixing test function.

franklx:
Test function fixed.

jknack:
Cool, are you done?

franklx:
Yes, it works and passes all tests.
Anyway it can be good if you take a closer look to the diff... I worked on quite sensitive code.

jknack:
Will do. Thanks for contributing!

PR #1598, opened 2020-03-24
Add phase option to openapi execution config

walkman90:


jknack:
We should add a note about the default phase: process-classes and when/why you might want to use a different phase (like. large projects)

walkman90:
@jknack pls check, feel free to change it whatever you like

Issue #1692, opened 2020-05-05
jooby2.x io.jooby.Usage: Router not found 


j2eekingcn:
hello,
I have set according to the documentation

	<plugin>
				<artifactId>maven-compiler-plugin</artifactId>
				<version>3.8.1</version>
				<configuration>
    				<forceJavacCompilerUse>true</forceJavacCompilerUse>
					
					<annotationProcessorPaths>
						<path>
						    <groupId>org.projectlombok</groupId>
						    <artifactId>lombok</artifactId>
						    <version>1.18.12</version>
						</path>
						<path>
							<groupId>io.jooby</groupId>
							<artifactId>jooby-apt</artifactId>
							<version>2.8.1</version>
						</path>
					</annotationProcessorPaths>
				</configuration>
			</plugin>
			<plugin>
				<groupId>io.jooby</groupId>
				<artifactId>jooby-maven-plugin</artifactId>
				<version>2.8.1</version>
				<configuration>
					<mainClass>${application.class}</mainClass>
					<restartExtensions>conf,properties,class</restartExtensions>
					<compileExtensions>java,kt</compileExtensions>
					<port>8080</port>
				</configuration>
			</plugin>

But after I modified the code
Console output error

io.jooby.Usage[m: [1;31mRouter not found: `cn.ef.modules.sys.action.DownloadAct`. Make sure Jooby annotation processor is configured properly.
For more details, please visit: https://jooby.io/usage#annotation-processing-tool-router-not-found[m
    [1mat[m io.jooby.Usage.apt ([1mUsage.java:62[m)
    [1mat[m io.jooby.Usage.mvcRouterNotFound ([1mUsage.java:41[m)
    [1mat[m io.jooby.Jooby.lambda$null$3 ([1mJooby.java:323[m)
    [1mat[m java.util.Optional.orElseThrow ([1mOptional.java:290[m)
    [1mat[m io.jooby.Jooby.lambda$mvc$4 ([1mJooby.java:323[m)
    [1mat[m java.util.Optional.orElseGet ([1mOptional.java:267[m)
    [1mat[m io.jooby.Jooby.mvc ([1mJooby.java:320[m)
    [1mat[m io.jooby.Jooby.mvc ([1mJooby.java:310[m)

j2eekingcn:
[...]

Issue #1390, opened 2019-09-20
jooby-apt: MVC annotations with intellij partial build


ajcamilo:
Intellij partially builds, so not all classes are "processed". Only when doing a full rebuild.

The problem is the `META-INF/services/io.jooby.MvcFactory` file is overwritten with only the "compiled" classes and not all MVC classes.

(Jooby 2.x)

jknack:
if that is the case... then we are lost. how can we fix this?

jerouris:
@ajcamilo Have you tried doing an initial full compilation from inside the IDE? IDEA keeps a kind of an index which is properly updated only when you do the compilation from inside it.

ajcamilo:
@jerouris yes, with the full compilation works, but when you change some of the MVC classes, the problem reappears (because the annotation processor only "sees" that class).

@jknack the MvcFactory is really needed? Can't you load the class `mvcClass + "API$Factory" in code? The services framework is really needed for "discovery", but this case you don't need to discover, because in the Jooby App, there will be an `mvc(MvcClass.class)` call.

PR #1767, opened 2020-06-07
Fixed invalid module name being used for MVC reflection fallback

jonjomckay:
This corrects the module name being used for the MVC reflection fallback being used when the module can't be found in the `ServiceLoader`.

I was running into #1390 even after the fix, and ended up finding the wrong class name suffix being used was the issue - now partial builds in IntelliJ work again :)


jonjomckay:
I think this might also fix #1692?

jknack:
Indeed will fix it. Thank you so much for this... I completely forgot about this "hack"

PR #1769, opened 2020-06-07
Added support for OpenAPI operating hiding annotations

### LINKED ISSUE #1768: OpenAPI annotations for hiding operations are ignored
If you have one or more methods annotated with either `@Hidden` or `@Operation(hidden = true)`, the `openapi` goal ignores these and sticks them into the OpenAPI output regardless. I saw there is an `excludes` configuration option for the goal, but I couldn't get it to work :(

jonjomckay:
Fixes #1768


jknack:
Excellent. Thank you

PR #1802, opened 2020-06-19
Add note about Maven GAV and modules

marcelstoer:
This is as much work in progress as the rest of the document. 

We have 20+ applications and a pile of custom modules. We are now evaluating the upgrade path. More guidance about the missing modules i.e. how to replace the ones not directly ported would definitely be needed.

### DIFF
--- a/docs/asciidoc/migration.adoc
+++ b/docs/asciidoc/migration.adoc
@@ -15,7 +15,7 @@ This is a **work in progress** document, if something is wrong or missing please
 |1.x|2.x
 |jooby-apitool| no real equivalent, use a combination of https://jooby.io/modules/openapi/[jooby-maven-plugin] and https://jooby.io/modules/openapi/#openapi-swagger-ui[jooby-swagger-ui]
 |jooby-hbv| n/a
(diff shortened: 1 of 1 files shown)


docs/asciidoc/migration.adoc
```diff
@@ -7,6 +7,18 @@ You will find here notes/tips about how to migrate from 1.x to 2.x.
...
 
+==== Maven coordinates
+`org.jooby` became `io.jooby`. Hence, use `<groupId>org.jooby</groupId>` for all dependencies.
+
+==== Modules
+|===
+|1.x|2.x
+|jooby-apitool| no real equivalent, use a combination of https://jooby.io/modules/openapi/[jooby-maven-plugin] and https://jooby.io/modules/openapi/#openapi-swagger-ui[jooby-swagger-ui]
+|jooby-hbv| n/a
+|jooby-lang-kotlin| n/a
```
jknack, line 18: Thanks. lang-kotlin is part of core now. You only need to add the kotlin dependencies for using it

marcelstoer: Yep, I just added a note.

PR #1841, opened 2020-07-09
Extension: commons-email from Jooby 1.x

imeszaros:
Almost a 1:1 port of `commons-email` module from Jooby 1.x

The only change is that only the `charset` is configured by default, and it is taken directly from `application.getConfig()`.

I created a `defaults(Config appConfig)` method to populate the default config for clarity. This method also sets an `originDescription` so the source of default configuration can be easily identified.


imeszaros:
@jknack do you have any idea why the checks are failing? My cloned repo all passed. Local build also works.

jknack:
looks at logs, there is a project version issue: 2.x branch is at 2.9.0-SNAPSHOT while your is at some 2.8.x version

imeszaros:
Thanks for the hint, rebased it and it now passes 👍

Issue #1345, opened 2019-07-17
ow2.asm Problem with Java 11


tobilebonk:
Hi :)

I encountered a problem while running Jooby 1.x under Java 11:

Environment:
```
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)
```

Stacktrace:
```
ERROR [atMyClass] An error occurred while starting the application:
java.lang.IllegalStateException: Can't read class: [anotherClass]
	at org.jooby.internal.RouteMetadata.extractMetadata(RouteMetadata.java:273)
	at com.google.common.cache.CacheLoader$FunctionToCacheLoader.load(CacheLoader.java:165)
	at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3528)
	at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2277)
	at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2154)
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2044)
	at com.google.common.cache.LocalCache.get(LocalCache.java:3952)
	at com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:3974)
	at com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache.java:4958)
	at com.google.common.cache.LocalCache$LocalLoadingCache.getUnchecked(LocalCache.java:4964)
	at org.jooby.internal.RouteMetadata.lambda$md$0(RouteMetadata.java:259)
	at org.jooby.funzy.Throwing.fn(Throwing.java:2423)
	at org.jooby.funzy.Throwing.access$400(Throwing.java:39)
	at org.jooby.funzy.Throwing$Supplier.get(Throwing.java:145)
	at org.jooby.funzy.Try.apply(Try.java:691)
	at org.jooby.internal.RouteMetadata.md(RouteMetadata.java:259)
	at org.jooby.internal.RouteMetadata.startAt(RouteMetadata.java:254)
	at org.jooby.internal.mvc.MvcRoutes.lambda$routes$2(MvcRoutes.java:288)
	at java.base/java.util.TimSort.countRunAndMakeAscending(TimSort.java:355)
	at java.base/java.util.TimSort.sort(TimSort.java:220)
[...]

PR #1849, opened 2020-07-14
Asm support for java 11

tobilebonk:
This PR refers to issue https://github.com/jooby-project/jooby/issues/1345 , which I posted about a year ago.

We were hoping that the issue above was resolved with 1.6.4 and revisited migrating our software to JAVA 11. Unfortunately we still got an ASM error (see issue above) with Jooby Version 1.6.7.

After some debugging we figured out that including the latest ASM dependency would not solve our problem as the ClassVisitors defined by jooby would always use ASM5. JAVA 11 however requires ASM7.

We therefore suggest to create the ClassVisitors using ASMv7. We also bumped the ASM version to the latest 7.x version.

We would really appreciate seeing that change in the next Jooby 1.x maintenance release :)

Greetings
Tobi


jknack:
Thanks for your PR. Jooby 1.6.8 has been released

PR #1861, opened 2020-07-19
Incremental annotation processing support with Gradle

### LINKED ISSUE #1860: Possible support for incremental annotation processing with Gradle (2.x)
Gradle supports incremental compilation _and_ incremental annotation processing. The prerequisite is that an annotation processor must report whether it's compatible with incremental processing or not. If an annotation processor is not compatible, or does not report that it is, when one annotated source file is changed, the __whole project__ must be recompiled. This is especially annoying if building in the IDE is delegated to the build tool (by default now in IDEA).

According to [the Gradle documentation](https://docs.gradle.org/4.7/userguide/java_plugin.html#_making_an_annotation_processor_incremental) processors can be categorized as __isolationg__ or __aggregating__ processors. After a brief look on the source code, my assumption is that `jooby-apt` could be an isolating processor. However, there are two requirements which currently arent met by `jooby-apt`:

- Can’t write resources
- Must provide exactly one originating element for each file generated with the Filer API

The first one is because the processor creates a service descriptor file used to load the MVC factories at runtime. My question would be here that what advantages come with using a service loader instead of instantiating with reflection? MVC extensions are installed at bootstrapping, so I guess it only could affect the startup performance?

The second thing should be easy to fix, the controller class could be provided as originating element for each generated module.

imeszaros:
Closes #1860

I found a good compromise. Now there is an annotation processor argument called `jooby.incremental` which toggles incremental processing support.

If `true`, `JoobyProcessor` reports to Gradle that it is an isolating processor and does not write the service descriptor file. Otherwise it works just like before.

The default settings is `false`. I extended the docs and cli also.

In case this PR gets merged, another PR could be created to list Jooby as a compatible annotation processor on Gradle's website: https://github.com/gradle/gradle/blob/master/subprojects/docs/src/docs/userguide/jvm/java_plugin.adoc#state-of-support-in-popular-annotation-processors


jknack:
can't we write the meta-inf file while running in incremental mode?

That's odd.

imeszaros:
Yes it is..

In Gradle > 6.0 you can actually write resource files using the Filer API, but the location must be one of `CLASS_OUTPUT`, `SOURCE_OUTPUT`, or `NATIVE_HEADER_OUTPUT`. Which is the case for Jooby fortunately.

The drawback is that you need to change it to be an "aggregating" processor because you need to provide __all__ originating elements to the resource file creation too.

With an aggregating processor Gradle will always recompile all of the generated classes, which is still better then recompiling everything.

I added another commit to change to this solution. I removed the doc entry and changed to be enabled by default.

imeszaros:
Maybe we could have another processor option like `jooby.serviceDescriptors` (by default `true`):
- `true`: generates metafile and fallback to __aggregating__ mode
- `false`: does __not__ generate metafile and report __isolating__ mode

jknack:
> Maybe we could have another processor option like jooby.serviceDescriptors (by default true):
true: generates metafile and fallback to aggregating mode
false: does not generate metafile and report isolating mode

I like this option better. We should add some doc notes too indicating how this option affects the gradle incremental annotation processor: recompile 1 vs all controllers. Also adding a note about when we use/fallback to reflection

imeszaros:
I pushed the changes above.

imeszaros:
Replaced by #1875

PR #1940, opened 2020-08-24
Ability to access the current context via the service registry

### LINKED ISSUE #1937: Accessing the Context via DI
For various reasons we need to be able to get the `Context` object of the current request via DI / Service Registry.

We're using the Guice extension and our current solution is another custom extension that wraps the Guice extension and creates the request scope as in 1.x and registers the `Context` in it.

Now I came up with a much simpler (and DI framework independent) solution:

```java
// install before any route:

public class ContextAsService implements Extension {

    private static final ThreadLocal<Context> provider = new ThreadLocal<>();

    @Override
    public void install(@Nonnull Jooby application) {
        application.getServices().putIfAbsent(Context.class, (Provider<Context>) provider::get);
        application.before(provider::set);
    }
}
```

My question would be, is there any better / preferred / official way to achieve this?

imeszaros:
Closes #1937

I wrote a simple test case and discovered a problem. With Jetty it works fine but fails with Netty and Undertow (strange because I did my initial tests with Netty, and it was working fine...).

The problem is that Netty and Undertow dispatches the execution to a worker thread. The thread local is initialized by the event loop thread, but then the service registry is queried by the worker thread -> not found.

Do you have an idea how to resolve this?

### DIFF
--- a/modules/jooby-jetty/src/main/java/io/jooby/internal/jetty/JettyContext.java
+++ b/modules/jooby-jetty/src/main/java/io/jooby/internal/jetty/JettyContext.java
@@ -557,8 +557,8 @@ public class JettyContext implements DefaultContext {
   @Nonnull @Override public Context onComplete(@Nonnull Route.Complete task) {
     if (listeners == null) {
       listeners = new CompletionListeners();
-      listeners.addListener(task);
     }
+    listeners.addListener(task);
     return this;
   }
 
--- a/docs/asciidoc/context.adoc
+++ b/docs/asciidoc/context.adoc
@@ -1,6 +1,46 @@
 == Context
 
-A javadoc:Context[Context] allows you to interact with the HTTP Request and manipulate the HTTP Response
+A javadoc:Context[Context] allows you to interact with the HTTP Request and manipulate the HTTP Response.
+
+In most of the cases you can access the context object as a parameter of your route handler:
+
+.Java
+[source, java, role="primary"]
+----
+{
+  get("/", ctx -> { /* do important stuff with variable 'ctx'. */ });
+}
+----
+
+.Kotlin
+[source, kotlin, role="secondary"]
+----
+{
+  get("/") { /* variable 'it' holds the context now. */ }
+}
+----
+
+If you need to access the context via the service registry or dependency injection, you need to
+explicitly request the registration of it as a service with the following invocation:
+
+.Java
+[source, java, role="primary"]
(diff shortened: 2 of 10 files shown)


jknack:
Try with `InheritableThreadLocal` but not sure... it worked before bc you wrote a filter which run the target thread

imeszaros:
I added a new internal api, `ExecutionCallback`. Not sure about the name though. The point is that it's called as soon as the context is passed to the final processing thread.

What do you think?

jknack:
yea, concept is right. what if we move existing ContextInitializer after thread dispatching? Like you did with ExecutionCallback. This way we don't need both...thoughts?

imeszaros:
Yes i was thinking about this too, but wouldn't that mess up with `HiddenMethodInitializer` and or `ContextInitializer#PROXY_PEER_ADDRESS`? They can now change which route matches with the context.

imeszaros:
Another possibility would be to reuse `ContextInitializer` and change the methods from:

```
addInitializer(ContextInitializer)
removeInitializer(ContextInitializer)
addExecutionCallback(ExecutionCallback)
removeExecutionCallback(ExecutionCallback)
```

to:

```
addPreDispatchInitializer(ContextInitializer)
removePreDispatchInitializer(ContextInitializer)
addPostDispatchInitializer(ContextInitializer)
removePostDispatchInitializer(ContextInitializer)
```

In the `Pipiline` class i wanted to use the `decorate` method to wrap the handler, but I noticed there is one single case it is not used, with `io.jooby.internal.Pipeline#kotlinJob`. Is this intentional or is it missing?

jknack:
Probably missing. Yea let's do that for now pre/post context initializer.

imeszaros:
I've done the changes. I also added separate commits for three more or less related bugs I discovered.

modules/jooby-jetty/src/main/java/io/jooby/internal/jetty/JettyContext.java
```diff
@@ -557,8 +557,8 @@ private Context sendStreamInternal(@Nonnull InputStream in) {
   @Nonnull @Override public Context onComplete(@Nonnull Route.Complete task) {
     if (listeners == null) {
       listeners = new CompletionListeners();
-      listeners.addListener(task);
     }
+    listeners.addListener(task);
```
jknack, line 561: Good catch!

Issue #1920, opened 2020-08-14
TransactionalRequest will make all router slower 10x，can it filter by attribute(@transactional)?


kindywu:
TransactionalRequest make all router slower，can it filter by attribute(@transactional)

example:decorator(new TransactionalRequest("transactional"));

jknack:
we could try to add something like that. but most application I wrote what I do is to put transactional where I know all the routes behind it are transactional:

```java

{
   // non transactional
   assets("...")
   ...

   // transactional
  decorator(new TransactionalRequest());

  ....
}
```

imeszaros:
Nobody misses the good old `UnitOfWork` concept? :) Our team does not like the transactional request concept, and we implemented the `UnitOfWork` from 1.x on our own. Is there a particular reason it's not the part of the 2.x module?

jknack:
isn't the same as transactional request we have today? or what did you miss from 1.x?

imeszaros:
In 1.x you could do something like this:

```
get('/do-something', req -> {
    req.require(UnitOfWork.class).apply(session -> {
        // do something
    });

    // and now something different
    req.require(UnitOfWork.class).apply(session -> {
        // do something else in an entirely new session/transaction
    });
});
```

But in 2.x, as far as I know, if you want the same functionality (i.e. use multiple sessions in the same request) you need to manually fetch and close the session, handle transaction etc.. I know it's not a very common use-case, but we really liked this idiom in 1.x.

jknack:
I see yea, don't think lot of people use it, but we could add it back for Hibernate (jdbi already have something similar). Not 100% sure about the name: `UnitOfWork` probably `SessionHandler`/`EntityManagerHandler`, others?

imeszaros:
`EntityManagerHandler` sounds OK to me, but `UnitOfWork` is a slightly more descriptive imho, though it does not say what kind of work it is about 😄 

There are a couple slight changes we made in our implementation, what do you think about them?
- It provides `EntityManager` instead of `Session` (we would like to use as little Hibernate specific features as possible.)
- It does not declare that it throws any exceptions (uses `sneakyThrow()`)
- does not support nesting anymore: this feature led to some confusion and misuse and therefor our implementation throws an exception if you try to instantiate while another session is open.

jknack:
Sounds perfect. I said *Handler bc that is the name in Jdbi. I would like t add couple more feature:

[...]

PR #1944, opened 2020-08-25
UnitOfWork port from 1.x

imeszaros:
See #1920

### DIFF
--- a/docs/asciidoc/modules/hibernate.adoc
+++ b/docs/asciidoc/modules/hibernate.adoc
@@ -188,6 +188,106 @@ Otherwise, you are going to see `LazyInitializationException`.
 There is a javadoc:hibernate.SessionRequest[] decorator that works identically but leaves transaction
 management to you, so no transaction is started/committed or rollback during a HTTP request.
 
+=== UnitOfWork
+
+Another way of managing the lifecycle of an EntityManager and transactions is working with an instance
+of `UnitOfWork`. You may acquire one from the service registry or inject it via DI.
+
+Usage:
+
+.UnitOfWork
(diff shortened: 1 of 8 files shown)


imeszaros:
Not really but i have no better candidate... maybe we could stick with UnitOfWork.. :) If you like this how it is now i'll write some docs later (once the name is final).

imeszaros:
I changed the name back to `UnitOfWork` and added docs.

jknack:
What if we copy (more or less) https://docs.spring.io/spring/docs/2.5.x/javadoc-api/org/springframework/orm/jpa/JpaTemplate.html and https://docs.spring.io/spring/docs/2.5.x/javadoc-api/org/springframework/orm/jpa/JpaCallback.html from Spring?

imeszaros:
You mean the name only, or the code/behavior too?

I can do that, but these classes were deprecated as of Spring 3.1, they recommend using the `EntityManager` directly. My personal opinion is that your approach with Jooby to be mostly minimalist and as invisible as possible is really great thing. We should not hide/wrap the JPA API, just provide some convenience around it.

I think `UnitOfWork` in 1.x was cool, we successfully brought several (mid to big) projects to life using it. The only feature we missed is the ability to handle multiple transactions, and the inconvenience that it declared to throw exception, these things i've tried to address in this PR.

Maybe it would be good to hear other hibernate user's opinion about this too.

If you wish I could try to bring these classes in instead, but in this case our team would probably stick with our custom extension.

jknack:
Yea, good points. Doubt any one else is going to provide feedback :D

For now, let's merge it.

Thank you.

PR #1955, opened 2020-08-30
Metrics ported from 1.x

imeszaros:
Metrics module ported from 1.x

I have three questions to clear:
- In the 1.x code argument `durationFactor` in method `io.jooby.metrics.MetricHandler#meter` is not used. Is it missing from somewhere or is it not actually needed?
- I wanted to keep the possibility to use DI to instantiate meters or health checks, therefore `MetricsModule` does half of the initialization in an `onStarted` callback. Is this approach OK (it's working of course) or should it be a lateinit extension?
- In 1.x routes could be selected by calling a config method with a path pattern. This can't be done in 2.x as far as I know, so I realized this functionality with a decorator. Is this the preferred way?


modules/jooby-metrics/src/main/java/io/jooby/metrics/MetricsModule.java
```diff
@@ -0,0 +1,277 @@
...
+
+    ServiceRegistry registry = application.getServices();
+
+    registry.putIfAbsent(MetricRegistry.class, metricRegistry);
+    registry.putIfAbsent(HealthCheckRegistry.class, healthCheckRegistry);
+
+    final Set<Reporter> reporters = new HashSet<>();
+
+    application.onStarted(() -> {
+      metrics.forEach((name, instance) -> {
```
jknack, line 239: Instances should not run on application started. We already have them so it is safe to do it on boot/init.

modules/jooby-metrics/src/main/java/io/jooby/metrics/MetricsModule.java
```diff
@@ -0,0 +1,277 @@
...
+        registry.putIfAbsent(ServiceKey.key(Metric.class, name), instance);
+        metricRegistry.register(name, instance);
+      });
+
+      metricClasses.forEach((name, clazz) -> {
+        Metric instance = application.require(clazz);
+        registry.putIfAbsent(ServiceKey.key(Metric.class, name), instance);
```
jknack, line 246: Classes must run on application started (after DI is ready) but there is no sense to add them in registry. Lifecycle of them are handle by DI framework.

modules/jooby-metrics/src/main/java/io/jooby/metrics/MetricsModule.java
```diff
@@ -0,0 +1,277 @@
...
+        metricRegistry.register(name, instance);
+      });
+
+      healthChecks.forEach((name, instance) -> {
```
jknack, line 250: Same comment about instances vs clsses on health checks.

modules/jooby-metrics/src/main/java/io/jooby/metrics/MetricsModule.java
```diff
@@ -0,0 +1,277 @@
...
+   * @return This metrics module.
+   */
+  public MetricsModule reporter(final Function<MetricRegistry, Reporter> callback) {
+    return reporter((registry, conf) -> callback.apply(registry));
+  }
+
+  @Override
+  public void install(@Nonnull Jooby application) {
+    MetricHandler metricHandler = new MetricHandler();
+    application.route("GET", this.pattern + "/metrics", metricHandler);
```
jknack, line 225: Let's do: `application.get(...)`

modules/jooby-metrics/src/main/java/io/jooby/internal/metrics/NoCacheHeader.java
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright 2014 Edgar Espina
+ */
+package io.jooby.internal.metrics;
+
+import io.jooby.Context;
+
+public class NoCacheHeader {
+
+  public static void add(Context ctx) {
+    ctx.setResponseHeader("Cache-Control", "must-revalidate,no-cache,no-store");
```
jknack, line 13: I don't like a class for adding a header. Can we add the header value as constant somewhere? Also, not sure we need it.

modules/jooby-metrics/src/main/java/io/jooby/metrics/MetricHandler.java
```diff
@@ -0,0 +1,193 @@
...
+        metrics.put("meters", meters);
+      }
+      Map<String, Object> timers = timers(registry.getTimers(filter), rateUnitLabel, rateFactor,
+          durationUnitLabel, durationFactor, showSamples);
+      if (timers.size() > 0) {
+        metrics.put("timers", timers);
+      }
+
+      // send
+      ctx.setResponseCode(StatusCode.OK);
```
jknack, line 87: no need to set `OK`

modules/jooby-metrics/src/main/java/io/jooby/metrics/MetricHandler.java
```diff
@@ -0,0 +1,193 @@
...
+  private static Map<String, Object> meters(final SortedMap<String, Meter> timers,
+      final String rateUnit, final double rateFactor, final String durationUnit,
+      final double durationFactor) {
+    Map<String, Object> result = new TreeMap<>();
+    timers.forEach((name, timer) -> result.put(name,
+        meter(timer, rateUnit, rateFactor, durationUnit, durationFactor)));
+    return result;
+  }
+
+  private static Map<String, Object> meter(final Metered meter, final String rateUnit,
```
jknack, line 175: I don't remember, but if we don't use it.. let's remove `durationFactor`

modules/jooby-metrics/src/main/java/io/jooby/metrics/MetricsModule.java
imeszaros: I see the point. Then maybe we should neither add the user code provided instances as well? Their life cycle should be handled by the user code.

jknack: Yea, but for DI usage what you did here is more clear and we make sure we request the object at time DI is ready. It is like more "safe". We need just these lines:

```
Metric instance = application.require(clazz);
metricRegistry.register(name, instance);
```

jknack:
There are some checkstyle errors, can you check?

PR #1957, opened 2020-08-30
Add @Transactional annotation to jooby-hibernate

### LINKED ISSUE #1920: TransactionalRequest will make all router slower 10x，can it filter by attribute(@transactional)?

imeszaros:
Closes #1920 

Doc page is coming once the API is settled.

### DIFF
--- a/modules/jooby-hibernate/src/main/java/io/jooby/hibernate/TransactionalRequest.java
+++ b/modules/jooby-hibernate/src/main/java/io/jooby/hibernate/TransactionalRequest.java
@@ -5,6 +5,7 @@
  */
 package io.jooby.hibernate;
 
+import io.jooby.Context;
 import io.jooby.Route;
 import io.jooby.ServiceKey;
 import io.jooby.SneakyThrows;
@@ -51,6 +52,8 @@ public class TransactionalRequest implements Route.Decorator {
 
   private ServiceKey<SessionFactory> key;
 
+  private boolean enabledByDefault = true;
+
   /**
    * Creates a new transactional request and attach the to a named session factory.
    *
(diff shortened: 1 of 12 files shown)


modules/jooby-hibernate/src/main/java/io/jooby/hibernate/TransactionalRequest.java
```diff
@@ -76,28 +79,63 @@ public TransactionalRequest(@Nonnull SessionRequest sessionRequest) {
...
   }
+
+  private boolean isTransactional(Context context) {
+    Object attribute = context.getRoute().attribute(Transactional.ATTRIBUTE);
+
+    if (attribute == null) {
+      return enabledByDefault;
+    }
+
+    if (!(attribute instanceof Boolean)) {
```
jknack, line 134: Let's invert here these lines:

```
if (attribute instanceof Boolean) {
   return attribute;
}

throw new IllegalArgumentException("Invalid value for route attribute "
          + Transactional.ATTRIBUTE + ": " + attribute);

```

imeszaros:
I've done the changes. I put the `isTransactional` method to the `Context` to avoid duplication. Now that the annotation is in the core i think that's a good place for it. What do you think?

imeszaros:
Do you have an idea where? I was thinking of the `io.jooby.Route` class too because this has to do something with route attributes.

jknack:
yea there: Route. Also, did you try this with MVC? I think mvc saves simple class name in camel case: `transactional`.

imeszaros:
Ok! Yes, but i'll double check it one more time.

imeszaros:
You're right, it does not work with MVC since I moved the annotation to core, because the `jooby-apt` ignores annotations as attributes beginning with `io.jooby.annotations`.

Possible solutions:
- move to another package (then where?)
- whitelist it by name
- create a meta-annotation like `@RouteAttribute` which is then recognized and allowed by `jooby-apt`

jknack:
oh crap! 3 seems more flexible.

imeszaros:
Done. We forgot about ebean.. :) I added support for that too.

jknack:
Awesome!

PR #1985, opened 2020-09-08
Advanced cache control for AssetHandler

imeszaros:
We used `AssetHandler` to serve a single page app and it does it great. However we would like to specify alternative cache header config for specific assets. We also need to explicitly disable caching completely for some assets. Currently we use our own asset handler implementation for this.

I created this PR as a basis of discussion for a possible feature for the above functionality in Jooby.

The API could be something like this:

```java
new AssetHandler(source)
        .cacheControl(path -> path.endsWith("about.html"), cacheControl.setNoCache())
        .cacheControl("foo.js", cacheControl -> cacheControl.maxAge(3600).setETag(false))
```

The current changes are a proof of concept and are fully backwards compatible.

### DIFF
--- a/jooby/src/main/java/io/jooby/CacheControl.java
+++ b/jooby/src/main/java/io/jooby/CacheControl.java
@@ -1,0 +1,150 @@
+/**
+ * Jooby https://jooby.io
+ * Apache License Version 2.0 https://jooby.io/LICENSE.txt
+ * Copyright 2014 Edgar Espina
+ */
+package io.jooby;
+
+import java.time.Duration;
+import java.util.function.Function;
+
+/**
+ * Class allowing the fine tune the browser cache behavior for assets.
+ *
+ * @see AssetHandler
+ * @see AssetHandler#cacheControl(Function)
+ */
+class CacheControl {
+
+  /**
+   * Constant for the max-age parameter, when set, no {@code Cache-Control} header is generated.
+   *
+   * @see #setMaxAge(long)
+   */
+  public static final int UNDEFINED = -1;
+
+  /**
+   * Constant for the max-age parameter, when set, the {@code Cache-Control} header is set to {@code no-store, must-revalidate}.
+   *
+   * @see #setMaxAge(long)
+   */
+  public static final int NO_CACHE = -2;
+
+  private boolean etag = true;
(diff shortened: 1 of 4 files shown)


jknack:
Looks good. What if we use a simple: `Function<String, CacheControl>` function? With this:

1. Unify API (single method)
2. remove the map

imeszaros:
I like the unified API idea and removal of the map, but it could be then rather a `BiConsumer<String, CacheControl>`. This way the callback could receive the default cache settings, and don't need to instantiate a new class.

Otherwise we need to replace the `CacheControl` interface with a bean (probably extract `CacheParams` class from `AssetHandler` and add getters).

Which way should wo choose?

jknack:
I feel the function makes more sense bc we need a new instance per asset, right?

imeszaros:
That's right yes, one instance per asset, but the idea was that the `AssetHandler` creates that instance and passes it to the callback for modification, and the instance is initialized the same way as the `AssetHandler` itself was configured. The callback only changes it if some deviation from the defaults is needed.

But if you feel it's cleaner to have a function that always creates the config from scratch, therefore completely overriding the defaults in `AssetHandler`, that's also fine i think.

jknack:
In general I do prefer a functional approach (no side effect), just sometimes isn't possible.

I do suggest to make CacheControl a class with factory methods. Then we setup the asset handler with a default cache function and give us the default cache control settings.

When we need more we just override the default cache function with a custom one.

A functional approach eliminates the internal class CacheParams.

What do you think?

imeszaros:
To be honest I also like it better this way. 😄 I also added a constant for `-1` of max-age.

jooby/src/main/java/io/jooby/CacheControl.java
```diff
@@ -0,0 +1,150 @@
...
+import java.time.Duration;
+import java.util.function.Function;
+
+/**
+ * Class allowing the fine tune the browser cache behavior for assets.
+ *
+ * @see AssetHandler
+ * @see AssetHandler#cacheControl(Function)
+ */
+class CacheControl {
```
jknack, line 17: make it public

jknack (changes requested):
Looking good

PR #1996, opened 2020-09-14
Allow to resolve URLs with context = null

### LINKED ISSUE #1990: UrlResolverImpl should work with context = null
Pac4j's `SAML2Client` does a [weird thing](https://github.com/pac4j/pac4j/blob/e6a3fbc762b02c7950b1a76457aa4f5c062ef034/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java#L112) at initialization which will cause an NPE to be thrown because `jooby-pac4j`'s `UrlResolverImpl` is not prepared for the case that the argument `context` can be `null`.

I have not the slightest idea why is it like this but I could only overcome this issue by plugging in a custom implementation of `UrlResolver` that itself builds the full URL purely from the applications `Config` instead of using the `context`.

imeszaros:
May close #1990

I had no better idea than generating the URL from scratch, with code similar to `DefaultContext#getRequestPath(String)`

### DIFF
--- a/tests/src/test/java/io/jooby/Issue1990.java
+++ b/tests/src/test/java/io/jooby/Issue1990.java
@@ -1,0 +1,77 @@
+package io.jooby;
+
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.read.ListAppender;
+import io.jooby.internal.pac4j.UrlResolverImpl;
+import io.jooby.junit.ServerTest;
+import io.jooby.junit.ServerTestRunner;
+import io.jooby.pac4j.Pac4jContext;
+import org.junit.jupiter.api.Test;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class Issue1990 {
+
+  @Test
+  public void withoutContextRelative() {
+    Logger log = (Logger) LoggerFactory.getLogger(UrlResolverImpl.class);
+    ListAppender<ILoggingEvent> appender = new ListAppender<>();
+    appender.start();
+    log.addAppender(appender);
+
+    UrlResolverImpl resolver = new UrlResolverImpl();
+
+    assertEquals("/callback", resolver.compute("/callback", null));
+    assertFalse(appender.list.isEmpty());
+
+    ILoggingEvent event = appender.list.get(0);
+
+    assertEquals("Unable to resolve URL from path '{}' since no web context was provided." +
+        " This may prevent some authentication clients to work properly." +
+        " Consider explicitly specifying an absolute callback URL or using a custom url resolver.", event.getMessage());
+
+    assertEquals(Level.WARN, event.getLevel());
+    assertNotNull(event.getArgumentArray());
+    assertTrue(event.getArgumentArray().length > 0);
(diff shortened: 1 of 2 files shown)


jknack (commented):
I don't like to redo UrlResolver using static variables (from application). The current version works more dynamically (from context).

Think we have two options here:

- See what happens if just return null on null context
- Ask pac4j author(s) why we have a null there

no?

imeszaros:
I agree it's not the best thing.. a few notes though:

I got the idea from default context, it does it very similar, e.g. if `getHostAndPort()` returns `null` it falls back server options. The change I introduced does ignore the context, is uses it as before if available.

`SAML2Client` passes `null` there because that code runs at initialization and the context is not available. Unfortunately you can't just return `null` because it stores the returned value and later passes it to the IDP as callback URL.

Interestingly their default implementation simple returns the path if the context is `null`:
https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/http/url/DefaultUrlResolver.java

This was the first I tried but in this situation it generated an incomplete callback URL containing only the path, which I think not sufficient for the IDP.

I'll investigate it more.

imeszaros:
I found a discussion about this issue: https://github.com/pac4j/pac4j/pull/1442

TL;DR: the conclusion is that they recommend using absolute URLs in the first place. Sending in `null` as context is by design.

Therefore I would suggest to change Jooby's implementation to simply return the path itself when the context is `null`. Then we get the out of the box pac4j behavior instead of the NPE.

If this does not fit for someone (like us) it's fairly easy to plug in a custom resolver.

What do you think?

jknack:
> Sending in null as context is by design.

This doesn't make any sense.

What if we add a new factory method that:

- takes an absolute url (String)
- we add a nonnull check on null context saying it is not supported and must use the static url resolver

imeszaros:
> This doesn't make any sense.

I agree but that is what they state here: https://github.com/pac4j/pac4j/pull/1442#issuecomment-553881479

> In SAML, the callback URL put in the metadata is computed at the initialization with a `null` web context, so the `CallbackUrlResolver` is called which calls the `UrlResolver` but this latter does nothing as the web context is `null`.
 
and later:

> Indeed, it might not be possible to properly compute a relative URL at startup, this is something I must investigate.
> The SP metadata is the data you will give to your IdP and I strongly doubt it will work if there is a relative URL in the metadata.

I see your point and it seems logical, but there is one problem with it. Unfortunately the `UrlResolver` is registered in the `Clients` instance (note the plural), and is therefore shared between all clients. If you use multiple types of `Client`s then this change would force you to use the same absolute URL for all (losing all the nice dynamic logic you built in like proxy handling etc).

If you check for example the `FormClient` it calls the resolver properly in request scope:
https://github.com/pac4j/pac4j/blob/4a42f190266b1f53af056b9344df220b8d84117b/pac4j-http/src/main/java/org/pac4j/http/client/indirect/FormClient.java#L69

I proposed to follow their logic because then we would not end up with an exception and basically the user would get the same behavior like pac4j was used without the Jooby integration.

What about emitting a warning in the log if `context == null && !isAbsoulte(path)` and returning the `path` instead of NPE?

jknack:
> What about emitting a warning in the log if context == null && !isAbsoulte(path) and returning the path instead of NPE?

Let's go with this.

tests/src/test/java/io/jooby/Issue1990.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class Issue1990 {
+
+  @Test
+  public void withoutContextRelative() {
+    Logger log = (Logger) LoggerFactory.getLogger(UrlResolverImpl.class);
+    ListAppender<ILoggingEvent> appender = new ListAppender<>();
```
jknack, line 26: cool, first time I see this!

PR #1999, opened 2020-09-19
OpenAPI generator now supports routes() and Kotlin objects

### LINKED ISSUE #1986: OpenAPI Generation doesn't work using kotlin object
Even simple example of mvc controller and openapi generation doesn't work, [here](https://github.com/abnud1/jooby-openapi-mvc-demo) you find an example project

Snippet:

```kotlin
package org.example

import io.jooby.Kooby
import io.jooby.OpenAPIModule
import io.jooby.annotations.GET
import io.jooby.annotations.Path
import io.jooby.runApp

@Path("/")
object Hello{
    @GET
    fun hello(): String{
        return "Hello World"
    }
}

class App : Kooby({
    install(OpenAPIModule())
    mvc(Hello)
})
fun main(args: Array<String>){
    runApp(args,App::class)
}
```

### LINKED ISSUE #1987: OpenAPI Generation doesn't work with nested routes block
I have code like this: 
```kotlin
class App : Kooby({
    install(OpenAPIModule())    
    // some before filters
    get("/someroute"){
          // do something
    }
    routes {        
        // before filters specific to these controllers
        mvc(Controller1())
        mvc(Controller2())
        ...
    }
})
```
Controller1,Controller2,etc are kotlin classes which uses MVC style and they work perfectly.

The only problem is that The generated openapi.yaml only contains the outside `get` route, the mvc routes inside `routes` block isn't found at all at openapi.yaml

In [this](https://github.com/abnud1/jooby-openapi-mvc-demo) project you find a demonstration of this problem.

Right now this is a blocker to me from OpenAPI since I have some routes on which I don't want to apply some before filters that's why I'm isolating those routes outside the `routes` block.

Oh and putting `install(OpenAPIModule())` inside `routes` block doesn't work either.

imeszaros:
May close #1987 and may also close #1986

### DIFF
--- a/modules/jooby-openapi/src/main/java/io/jooby/internal/openapi/AnnotationParser.java
+++ b/modules/jooby-openapi/src/main/java/io/jooby/internal/openapi/AnnotationParser.java
@@ -28,6 +28,7 @@ import org.objectweb.asm.Type;
 import org.objectweb.asm.tree.AbstractInsnNode;
 import org.objectweb.asm.tree.AnnotationNode;
 import org.objectweb.asm.tree.ClassNode;
+import org.objectweb.asm.tree.FieldInsnNode;
 import org.objectweb.asm.tree.LdcInsnNode;
 import org.objectweb.asm.tree.LocalVariableNode;
 import org.objectweb.asm.tree.MethodInsnNode;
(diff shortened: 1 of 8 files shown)


jknack:
byte code rock star 😎

PR #2000, opened 2020-09-20
Cookie API now supports the SameSite parameter

### LINKED ISSUE #1998: Chrome cookie SameSite issue
When redirecting in Chrome, set-cookie will be ignored if SameSite is not set.
Please refer: [Chrome's Changes Could Break Your App: Prepare for SameSite Cookie Updates](https://blog.heroku.com/chrome-changes-samesite-cookie)
After review codes, I found maybe these classes need to be modified:

1. For every Context implementation, must set secure to true in cookie if using https.
    In my case, I am using NettyContext, so I must do this:
```
@Nonnull public Context setResponseCookie(@Nonnull Cookie cookie) {
    if (responseCookies == null) {
      responseCookies = new HashMap<>();
    }
    cookie.setPath(cookie.getPath(getContextPath()));
    cookie.setSecure(isSecure());   // <== Add this line
    responseCookies.put(cookie.getName(), cookie.toCookieString());
    setHeaders.remove(SET_COOKIE);
    for (String cookieString : responseCookies.values()) {
      setHeaders.add(SET_COOKIE, cookieString);
    }
    return this;
  }
```
2. In **Cookie.java**, modify toCookieString():
```
  public @Nonnull String toCookieString() {
    ...
    // Secure
    if (secure) {
      sb.append(";Secure;SameSite=None");
    }
    ....
    return sb.toString();
  }
```

imeszaros:
May close #1998

This PR contains the possible API changes. This should not change anything on the current behavior unless someone sets the parameter explicitly.

We should now decide whether to apply this on any of jooby's cookies (session, flash, pac4j maybe)?
- [DefaultContext](https://github.com/jooby-project/jooby/blob/d6fbe9b552501240719c5ee9c9fe698a059c8ad7/jooby/src/main/java/io/jooby/DefaultContext.java#L95)
- [SessionToken](https://github.com/jooby-project/jooby/blob/d6fbe9b552501240719c5ee9c9fe698a059c8ad7/jooby/src/main/java/io/jooby/SessionToken.java#L131)
- [WebContextImpl](https://github.com/jooby-project/jooby/blob/d6fbe9b552501240719c5ee9c9fe698a059c8ad7/modules/jooby-pac4j/src/main/java/io/jooby/internal/pac4j/WebContextImpl.java#L115)

In Play Framework they use `Lax` by default but it can be overridden from the config: https://github.com/playframework/playframework/pull/7103

pac4j's cookie API does not support it yet, so we should probably not add it by default, maybe allow to set it via configuration.

### DIFF
--- a/docs/asciidoc/context.adoc
+++ b/docs/asciidoc/context.adoc
@@ -557,16 +557,16 @@ a javadoc:Session[] but the lifecycle is shorter: *data is kept for only one req
 <3> Display an existing flash attribute `success` or shows `Welcome!`
 
 Flash attributes are implemented using an `HTTP Cookie`. To customize the cookie
-(its name defaults to `jooby.flash`) use the javadoc:Router[setFlashCookieTemplate, Cookie] method:
+(its name defaults to `jooby.flash`) use the javadoc:Router[setFlashCookie, Cookie] method:
 
 .Java
 [source,java,role="primary"]
 ----
 {
-  setFlashCookieTemplate(new Cookie("myflash").setHttpOnly(true));
+  setFlashCookie(new Cookie("myflash").setHttpOnly(true));
   
   // or if you're fine with the default name
-  getFlashCookieTemplate().setHttpOnly(true);
+  getFlashCookie().setHttpOnly(true);
 }
 ----
 
@@ -574,10 +574,10 @@ Flash attributes are implemented using an `HTTP Cookie`. To customize the cookie
 [source,kotlin,role="secondary"]
 ----
 {
-  flashCookieTemplate = Cookie("myflash").setHttpOnly(true)
+  flashCookie = Cookie("myflash").setHttpOnly(true)
 
   // or if you're fine with the default name
-  flashCookieTemplate.isHttpOnly = true
+  flashCookie.isHttpOnly = true
 }
 ----
 
--- a/docs/asciidoc/context.adoc
+++ b/docs/asciidoc/context.adoc
@@ -556,19 +556,29 @@ a javadoc:Session[] but the lifecycle is shorter: *data is kept for only one req
 <2> Redirect to home page
 <3> Display an existing flash attribute `success` or shows `Welcome!`
 
-Flash attributes are implemented using an `HTTP Cookie`. To customize the cookie name
-(defaults to `jooby.flash`) use the javadoc:FlashScope extension:
+Flash attributes are implemented using an `HTTP Cookie`. To customize the cookie
+(its name defaults to `jooby.flash`) use the javadoc:Router[setFlashCookieTemplate, Cookie] method:
 
 .Java
 [source,java,role="primary"]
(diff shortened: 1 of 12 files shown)


jknack (approved):

imeszaros:
I've analyzed the above three topics:

- Flash scope: user is not able to specify `SameSite` with current API
  - we could have a setter in router like `setFlashCookieSameSite` similar to the existing `setFlashCookie`
  - or we could have an override like `setFlashCookie(Cookie template)` where the user could set a template for the flash cookie
  - or we could have a config key like `jooby.flash.cookie` where the user can configure the cookie
- Session Cookie: the user can access `SessionToken#SID` so the `SameSite` can be set without any changes
- pac4J: until pac4j has built in support we can allow to configure it like `pac4j.cookies.sameSite`, WebContextImpl can get the config from `context -> router -> config`

What do you think?

jknack:
just realize documentation for flash attributes is wrong: https://jooby.io/#context-parameters-flash. It mentioned FlashScope but that doesn't exist in 2.x

Let's add `setFlashCookie(Cookie template)` and deprecate and refactor `setFlashCookie(String name)` to use the new method.

imeszaros:
I added a commit for the flash cookie changes and also fixed the docs. Unfortunately I had to name the method `get|setFlashCookieTemplate` because otherwise the getter could not be deprecated (signature collision).

Should I do something with the pac4j module and the `SID` cookie?

jknack:
Let's replace the getFlashCookie: String with the new and deprecate the setFlashCookie(String)

do we need anything for sid? There is a cookie option

> pac4J: until pac4j has built in support we can allow to configure it like pac4j.cookies.sameSite, WebContextImpl can get the config from context -> router -> config

Go for it.

imeszaros:
Bumped into a problem.. If I follow the existing pattern and use `Pac4jOptions` to provide the API for setting the `SameSite` value I need to inject the `Pac4jOptions` instance to `WebContextImpl`.

This is no big deal but there is one single public API in `Pac4jContext` class that has to be changed from:

```
  static @Nonnull Pac4jContext create(@Nonnull Context ctx) {
    return new WebContextImpl(ctx);
  }
```

to:

```
  static @Nonnull Pac4jContext create(@Nonnull Context ctx, @Nonnull Pac4jOptions options) {
    return new WebContextImpl(ctx, options);
  }
```

If this is not acceptable then I could get the value from the application config as I planned originally, but then it's different from every other setting..

jknack:
Go. to Pac4jModule and add Pac4jOptions to the service registry. Then from Pac4jContext ask Jooby context for them `ctx.require(Pac4jOptions.class`

does it work?

imeszaros:
I could easily test it with the dev login form since it sets a CSRF cookie. It works, thanks for the good suggestion!

jknack (approved):
Outstanding!

PR #2018, opened 2020-09-23
bump rxjava from io.reactivex.rxjava2 to io.reactivex.rxjava3, add metrics support to jooby-redis

aftershadow:
upgrade to rxjava3.0.6


jknack:
are you happy on how it works?

Honestly, thought no one used it.

Also, we had been discussing to move reactive to his own module (like we did in 1.x). This will helps to remove reflection from core around these types but also keep supporting rx 2.x.

aftershadow:
It is not ready for production use, and is still doing comparative testing and verification.
Considering that Rx2.x has entered maintenance mode, this means that there may be conflicts in future new features or components that rely on it. From the current version, the difference between 2.x and 3.x is still acceptable.
So, why not use the new Rx3.x?

jknack:
Yea, we should support 3.x. What I'm not sure is to drop 2.x while we do it

aftershadow:
Yes, you are right.
Thanks

aftershadow:
MetricRegistry metricRegistry = new MetricRegistry();
        HealthCheckRegistry healthCheckRegistry = new HealthCheckRegistry();
        install(new RedisModule("foo").withComputationThreadPoolSize(4).withIOThreadPoolSize(4).metricRegistry(Duration.ofSeconds(10),metricRegistry));
        install(new RedisModule("bar").metricRegistry(Duration.ofSeconds(10),metricRegistry));
        install(new HikariModule()
                .metricRegistry(metricRegistry)
                .healthCheckRegistry(healthCheckRegistry));
        install(new MetricsModule(metricRegistry,healthCheckRegistry)
                .threadDump()
                .ping()
        );

jknack:
upgrade to rx 3 and reactor is done in 3.x branch

PR #2019, opened 2020-09-24
Ability to match request locale and use it for templates

### LINKED ISSUE #2006: Jooby 2.x Pebble Module i18n support
According to the Pebble documentation, i18n is handled via the evaluate() method:
```Every PebbleTemplate is assigned a default locale from the PebbleEngine. At the point of evaluation, this locale can be changed with an argument to the evaluate(...) method of the individual template.```

However, currently the Jooby Pebble module does not pass any locale information to this method:
```@Override
  public String render(Context ctx, ModelAndView modelAndView) throws Exception {
    PebbleTemplate template = engine.getTemplate(modelAndView.getView());
    Writer writer = new StringWriter();
    Map<String, Object> model = new HashMap<>(ctx.getAttributes());
    model.putAll(modelAndView.getModel());
    template.evaluate(writer, model);
    return writer.toString();
  }```

It would be nice to have built-in support for this, e.g. via a setLocale() method on the request context or similar.

### PR DESCRIPTION
**Pull request by imeszaros**
May close #2006

I need to write test cases and test it more, but what do you think?

I didn't find any way to set the locale for rocker and handlebars.

### DIFF
--- a/modules/jooby-thymeleaf/src/main/java/io/jooby/internal/thymeleaf/ThymeleafTemplateEngine.java
+++ b/modules/jooby-thymeleaf/src/main/java/io/jooby/internal/thymeleaf/ThymeleafTemplateEngine.java
@@ -37,12 +37,7 @@ public class ThymeleafTemplateEngine implements io.jooby.TemplateEngine {
     // Locale:
     Locale locale = modelAndView.getLocale();
     if (locale == null) {
-      // keep compatibility with previous versions
-      locale = (Locale) model.getOrDefault("locale", null);
-
-      if (locale == null) {
-        locale = ctx.locale();
-      }
+      locale = ctx.locale();
     }
 
     Context context = new Context(locale, model);
--- a/modules/jooby-thymeleaf/src/main/java/io/jooby/internal/thymeleaf/ThymeleafTemplateEngine.java
+++ b/modules/jooby-thymeleaf/src/main/java/io/jooby/internal/thymeleaf/ThymeleafTemplateEngine.java
@@ -35,7 +35,15 @@ public class ThymeleafTemplateEngine implements io.jooby.TemplateEngine {
     model.putAll(modelAndView.getModel());
 
     // Locale:
-    Locale locale = (Locale) model.getOrDefault("locale", Locale.getDefault());
+    Locale locale = modelAndView.getLocale();
+    if (locale == null) {
+      // keep compatibility with previous versions
+      locale = (Locale) model.getOrDefault("locale", null);
+
+      if (locale == null) {
+        locale = ctx.locale();
+      }
+    }
 
     Context context = new Context(locale, model);
     String templateName = modelAndView.getView();
--- a/tests/src/test/java/io/jooby/FeaturedTest.java
+++ b/tests/src/test/java/io/jooby/FeaturedTest.java
@@ -3058,6 +3058,32 @@ public class FeaturedTest {
     });
   }
 
+  @ServerTest
+  public void locales(ServerTestRunner runner) {
+    runner.define(app -> {
+      app.get("/locales", Context::locales);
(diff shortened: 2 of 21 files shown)

### CODE REVIEW

**Comment by jknack** [comment]
Looks very good.

File: tests/src/test/java/io/jooby/FeaturedTest.java [hunk 1]
```diff
@@ -3058,6 +3058,32 @@ public void regexpRoutingShouldIncludeDefaultValueWhenParamIsNotMatched(ServerTe
...
+      app.get("/locale", Context::locale);
+      app.setLocales(Arrays.asList(new Locale("en"),
+          new Locale("en", "GB"),
+          new Locale("de", "AT"),
+          new Locale("de", "CH"),
+          new Locale("fr"),
+          new Locale("hu")));
+    }).dontFollowRedirects().ready(client -> {
+      client.header("Accept-Language", "fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5")
+          .get("/locales", rsp -> assertEquals("[en, en_GB, de_AT, de_CH, fr, hu]", rsp.body().string()));
```
imeszaros, line 3074: If `*;q=0.5` is in the header, the list returned by `Locale.filter(...)` is not sorted by priority.... have no idea why.

tests/src/test/java/io/jooby/FeaturedTest.java
```diff
@@ -3058,6 +3058,32 @@ public void regexpRoutingShouldIncludeDefaultValueWhenParamIsNotMatched(ServerTe
...
+
+      client.header("Accept-Language", "fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7")
+          .get("/locales", rsp -> assertEquals("[fr, en, en_GB, de_AT, de_CH]", rsp.body().string()));
```
imeszaros, line 3077: Without the asterisk it works as expected...

imeszaros:
I added the tests. I think it's ok but there is a strange behavior I commented above. This language matching concept is not very straightforward for me.

imeszaros:
Ok I read through [RFC 4647](https://tools.ietf.org/html/rfc4647) and actually it's not mandatory that filter returns an ordered list:

>   If the language priority list (see Section 2.3) contains more than
   one range, the content returned is typically ordered in descending
   level of preference, but it MAY be unordered, according to the needs
   of the application or protocol.

So i read the source of `LocaleMatcher` found [here](https://github.com/frohoff/jdk8u-dev-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/classes/sun/util/locale/LocaleMatcher.java#L118), and indeed when one of the tags in the priority list is the asterisk (`*`), it just returns the input list as it is.

tests/src/test/java/io/jooby/FeaturedTest.java
imeszaros: It turned out that it's by design.

modules/jooby-thymeleaf/src/main/java/io/jooby/internal/thymeleaf/ThymeleafTemplateEngine.java
```diff
@@ -35,7 +35,15 @@ public ThymeleafTemplateEngine(TemplateEngine templateEngine, List<String> exten
     model.putAll(modelAndView.getModel());
 
     // Locale:
-    Locale locale = (Locale) model.getOrDefault("locale", Locale.getDefault());
+    Locale locale = modelAndView.getLocale();
+    if (locale == null) {
+      // keep compatibility with previous versions
+      locale = (Locale) model.getOrDefault("locale", null);
```
jknack, line 41: I doubt anyone use this and if someone does, let's force them to upgrade.

jknack:
just one change and we are ready.

PR #2022, opened 2020-09-27
Gradle: joobyRun now shutdown properly on Ctrl+C

### LINKED ISSUE #1947: Intellij IDEA can't stop jooby plugin completely
Using Intellij IDEA(2020.2) & gradle (6.6) & jooby plugin (latest version)
If I press the red square stop button, the joobyRun task failed to stop completely.
When I start joobyRun task again, it gives me "Address already in use: 8080" exception.
I have to kill the underline java process manunly

jknack:
- Shutdown hooks are not trigger by Gradle daemon, removed the shutdown hook
- We start a new thread to shutdown on Enter key (this still works), this same threads throws a NoSuchElementException when Gradle disconnect I/O
- We now block this thread either by I/O disconnectd or Enter key and run a shutdown
- Conclusion:
  - Ctrl+ C works as expected (normal shutdown)
  - IDE integration works as expected (normal shutdown)
  - In both cases, daemon can be reused in next run!!!

Fix #1947

### DIFF
--- a/modules/jooby-gradle-plugin/src/main/java/io/jooby/gradle/RunTask.java
+++ b/modules/jooby-gradle-plugin/src/main/java/io/jooby/gradle/RunTask.java
@@ -5,8 +5,16 @@
  */
 package io.jooby.gradle;
 
-import io.jooby.run.JoobyRun;
-import io.jooby.run.JoobyRunOptions;
+import java.io.File;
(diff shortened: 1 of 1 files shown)


imeszaros (approved):
Tested it and it works great for me in IDEA and CLI.

PR #2023, opened 2020-09-28
pac4j-module: Can't apply a client directly to a POST route. Fix #1972

### LINKED ISSUE #1972: pac4j-module: Can't apply a client directly to a POST route?
I am using the pac4j module to do JWT authentication on my API routes.  I currently have a POST endpoint that is at a concrete route (`/auth/register`), and I want to add a pac4j client to that route:
```
        install(
            Pac4jModule()
                .client("/auth/register") { conf ->
                    DirectBearerAuthClient(get() as JwtAuthenticator) // (using koin to inject JwtAuthenticator)
                }
  ...
```

... but I have found that the endpoint doesn't actually authenticate credentials unless it's a GET instead of a POST.  In debugging, I found this line:
https://github.com/jooby-project/jooby/blob/2.x/modules/jooby-pac4j/src/main/java/io/jooby/pac4j/Pac4jModule.java#L363

It looks to me that unless there are variables (or maybe wildcards?) in the path, the pac4j module assumes that it's a GET path.

It's entirely possible that I'm doing it wrong!  Should I only apply the pac4j client to a larger path with a wildcard?  Is there a trick to making it work?  Am I doing something stupid without realizing it?  :-P  Any pointers appreciated!

jknack:


imeszaros (approved):
It definitely fixes the issue, but is there any chance that other http methods needs to be covered too? I'm not sure what are the use cases when you register a client directly against a route instead of a decorator.

jknack:
> It definitely fixes the issue, but is there any chance that other http methods needs to be covered too?

Probably, but we won't extend this to any other method.

This is just for very simple use cases where direct client is necessary (no session, redirect in the middle) that is why GET and POST are ok. Filter/decorator is for more complex uses cases.

imeszaros:
> This is just for very simple use cases where direct client is necessary (no session, redirect in the middle) that is why GET and POST are ok. Filter/decorator is for more complex uses cases.

In this case i think it's good to go!

PR #2029, opened 2020-09-29
CLI: change unit test for MVC apps not to use MockRouter

### LINKED ISSUE #1917: MockRouter does not support MVC Coroutines
It seems like the CoroutineRouter might make a few different assumptions around not using the MVC mode in general, but the main one blocking us at the moment is the MockRouter not working properly.

I see this commit fixed this for the other mode by setting a new attribute: https://github.com/jooby-project/jooby/commit/8ed3364810a811844095fdf4182ecbf7b81d8c05

But it appears it wasn't implemented for the MVC side of things - am I understanding that correctly?

Thanks!

imeszaros:
May close #1917

### DIFF
--- a/modules/jooby-cli/src/main/java/io/jooby/cli/CreateCmd.java
+++ b/modules/jooby-cli/src/main/java/io/jooby/cli/CreateCmd.java
@@ -109,9 +109,9 @@ public class CreateCmd extends Cmd {
     String server;
     boolean stork = !gradle && this.stork;
     if (interactive) {
-      gradle = yesNo(ctx.readLine("Use Gradle (yes/No): "));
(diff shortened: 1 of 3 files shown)


jknack:
Thanks!, what if we provide a real unit test for mvc controller?

imeszaros:
I was considering it but not exactly clear how should it look like exactly.

jknack:
I looked at the generated code and it is just:

```
import io.jooby.annotations.*;

@Path("/")
public class Controller {

  @GET
  public String sayHi() {
    return "Welcome to Jooby!";
  }
}
```

So unit tests should be very simple...not sure if useful

imeszaros:
So just a simple test against the controller class, no jooby logic involved? ok I can do that.

jknack (approved):

PR #2054, opened 2020-10-15
Fix exception handling at startup time

### LINKED ISSUE #2053: question: Error handling at startup time
Hi. Why does the cause propagated here instead of the source exception? If at startup time I throw an exception with custom message `throw new IllegalArgumentException("message", cause)`, then the message will be lost.

https://github.com/jooby-project/jooby/blob/76f04546e45611a120e32d790d3de280c0f8df10/modules/jooby-utow/src/main/java/io/jooby/utow/Utow.java#L122

extbe:
Fixes #2053


jknack (approved):
Good catch! Thank you

PR #2060, opened 2020-10-20
Add Jooby Metrics to list of unsupported 2.x modules

marcelstoer:


jknack:
It is supported: https://jooby.io/modules/metrics/ !!!

marcelstoer:
Sorry, I missed that it was apparently released a month ago with v2.9.0.

PR #2078, opened 2020-10-26
SSL: configure protocol

jknack:
- Set default protocol to `TLSv1.2, TLSv1.3`
- If a listed protocol is not supported, it is ignored; however, if you specify a list of protocols, none of which are supported, an exception will be thrown.
- Turn off TLSv1 and TLSv1.1 in Jetty
- Fix date tests failure

### DIFF
--- a/docs/asciidoc/servers.adoc
+++ b/docs/asciidoc/servers.adoc
@@ -4,9 +4,9 @@ Jooby comes with three server implementations:
 
 - javadoc:Jetty[]
 - javadoc:Netty[]
-- javadoc:Undertow[]
+- javadoc:Utow[]
 
-Server are automatically registered based on their presence on the project classpath.
+Servers are automatically registered based on their presence on the project classpath.
 
 To use Jetty, add the dependency:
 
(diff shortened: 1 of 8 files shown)


docs/asciidoc/servers.adoc
```diff
@@ -1,12 +1,12 @@
 == Server
 
-Jooby comes with three server implementations:
+Jooby comes with three servers implementations:
```
imeszaros, line 3: Singular was correct here.

imeszaros (approved):
Thumbs up for disabling TLSv1 and TLSv1.1!

jooby/src/main/java/io/jooby/SslOptions.java
```diff
@@ -72,6 +87,8 @@
 
   private ClientAuth clientAuth = ClientAuth.NONE;
 
+  private List<String> protocol = Arrays.asList(TLS_V1_2, TLS_V1_3);
```
imeszaros, line 90: It may be better to list 1.3 first, if a server implementation takes care about the order.

imeszaros (approved):

jooby/src/main/java/io/jooby/SslOptions.java
jknack: Not sure if they care, but I'm reverting.

jknack:
updated with:

- review comments
- added unit test for parsing SslOptions.protocol from config
- some minor doc changes

PR #2089, opened 2020-10-29
add kafka module to jooby

kindywu:
/**
   * Creates a new kafka producer module using the <code>kafka.producer</code> property key.
   * This key must be present in the application configuration file, like:
   *
   * <pre>{@code
   *  kafka.producer.bootstrap.servers = "localhost:9092"
   *  kafka.producer.acks = "all"
   *  kafka.producer.retries = 0
   *  kafka.producer.key.serializer = "org.apache.kafka.common.serialization.StringSerializer"
   *  kafka.producer.value.serializer = "org.apache.kafka.common.serialization.StringSerializer"
   * }</pre>
   *
   * Creates a new kafka consumer module using the <code>kafka.consumer</code> property key.
   * This key must be present in the application configuration file, like:
   *
   * <pre>{@code
   *  kafka.consumer.bootstrap.servers = "localhost:9092"
   *  kafka.consumer.group.id = "group A"
   *  kafka.consumer.key.deserializer = "org.apache.kafka.common.serialization.StringDeserializer"
   *  kafka.consumer.value.deserializer = "org.apache.kafka.common.serialization.StringDeserializer"
   * }</pre>
   */

### DIFF
--- a/modules/jooby-kafka/src/main/java/io/jooby/kafka/KafkaModule.java
+++ b/modules/jooby-kafka/src/main/java/io/jooby/kafka/KafkaModule.java
@@ -1,0 +1,119 @@
+/**
+ * Jooby https://jooby.io
+ * Apache License Version 2.0 https://jooby.io/LICENSE.txt
+ * Copyright 2014 Edgar Espina
+ */
+package io.jooby.kafka;
+
+import com.typesafe.config.Config;
+import io.jooby.Extension;
+import io.jooby.Jooby;
+import io.jooby.ServiceRegistry;
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.clients.producer.KafkaProducer;
+
+import javax.annotation.Nonnull;
+import java.util.Properties;
+
+/**
+ * Redis module: https://jooby.io/modules/redis.
+ * <p>
+ * Usage:
+ *
+ * <pre>{@code
+ * {
+ *   install(new RedisModule());
+ *
+ *   get("/", ctx -> {
+ *     StatefulRedisConnection redis = require(StatefulRedisConnection.class);
+ *     // work with redis
+ *   });
+ * }
+ * }</pre>
+ * <p>
+ * application.conf:
+ *
+ * <pre>
+ *   redis = "redis://localhost:6379"
+ * </pre>
+ * <p>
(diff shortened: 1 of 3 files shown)


modules/jooby-kafka/src/main/java/io/jooby/kafka/KafkaModule.java
```diff
@@ -0,0 +1,156 @@
...
+ * </ul>
+ * <p>
+ * Alternative you can pass a redis URI:
+ * <pre>
+ *   install(new RedisModule("redis://localhost:6379"));
+ * </pre>
+ *
+ * @author edgar
+ * @since 2.8.5
+ */
```
imeszaros, line 56: Please replace the docs copied from Redis module with proper docs.

modules/jooby-kafka/src/main/java/io/jooby/kafka/KafkaModule.java
```diff
@@ -0,0 +1,156 @@
...
+
+    if (this.consumerProps != null) {
+      KafkaConsumer consumer = new KafkaConsumer(consumerProps);
+      application.onStop(consumer::close);
+      registry.putIfAbsent(KafkaConsumer.class, consumer);
+    }
+  }
+
+  private static Properties properties(final Config config, final String key) {
+    Properties props = new Properties();
```
imeszaros, line 148: You could throw an exception if `key` argument is `null`, since it means an improper configuration.

modules/jooby-kafka/src/main/java/io/jooby/kafka/KafkaModule.java
```diff
@@ -0,0 +1,156 @@
...
+    if (config.hasPath(key)) {
+      // dump
+      config.getConfig(key).entrySet().forEach(
+          e -> props.setProperty(e.getKey(), e.getValue().unwrapped().toString()));
+    }
```
imeszaros, line 153: You could throw an exception in an `else` block here, since this also means that configuration properties are missing.

imeszaros (changes requested):
Beside the minor things I commented I find this okay (though I'm not familiar with Kafka)!

jknack:
Going to merge and cleanup

PR #2101, opened 2020-11-08
Rocker: move from string builder to byte array

jknack:
- Replace the StringBuilder output with a byte buffer array
- Byte buffer need to be sized properly to reduce re-allocation and copy
- Default buffer size is 4k
- There is an option to reuse/recycle a byte buffer using a thread local mechanism
- Extra: micro optimization around incoming requests: Ignore HTTP body on GET requests

### DIFF
--- a/docs/asciidoc/modules/rocker.adoc
+++ b/docs/asciidoc/modules/rocker.adoc
@@ -94,3 +94,53 @@ import io.jooby.rocker.RockerModule
 
 <1> Install Rocker
 <2> Returns a rocker view
+
+=== Options
+
+Rocker uses a byte buffer to render a view. Default byte buffer size is `4k`. To change the buffer size:
+
+.Java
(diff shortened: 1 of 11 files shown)


docs/asciidoc/modules/rocker.adoc
```diff
@@ -94,3 +94,53 @@ import io.jooby.rocker.RockerModule
...
+----
+import io.jooby.rocker.RockerModule
+
+{
+  install(RockerModule().reuseBuffer(true)
+}
+----
+
+CAUTION: Use with caution due it creates a buffer per thread memory consumption might be high.
+This technique works when the number of available threads is slow enough
```
imeszaros, line 145: Just a typo: **low** enough.

imeszaros (approved):
Looks good to me!

PR #2108, opened 2020-11-11
Logging exceptions occurred at app initialization.

### LINKED ISSUE #2107: App-class initialization errors don't get logged
Using the following code:

```java
public class App extends Jooby {

  {
    install(new SomeModuleThrowingExceptionAtInstall());
    get("/", ctx -> "Welcome to Jooby!");
  }

    runApp(args, App::new);
  }
}
```

The exception thrown in the `install()` method of the module is not visible in the log because neither `Jooby.runApp()` or `Jooby.createApp()` does not catch any exceptions. It's only visible on the standard error output which is not ideal if the app is already deployed. We experienced this with the Hibernate module which throws an exception if there is a schema-validation error.

I suggest to catch the exceptions in `createApp()` and log them. At that time logging is already available.

imeszaros:
Closes #2107

### DIFF
--- a/jooby/src/main/java/io/jooby/Jooby.java
+++ b/jooby/src/main/java/io/jooby/Jooby.java
@@ -725,7 +725,9 @@ public class Jooby implements Router, Registry {
         log.info("Server stop resulted in exception", stopx);
       }
       // rethrow
-      throw new StartupException("Application startup resulted in exception", x);
+      throw x instanceof StartupException
+          ? (StartupException) x
+          : new StartupException("Application startup resulted in exception", x);
     }
   }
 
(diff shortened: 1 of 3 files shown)


jknack (approved):
Looks good, should we wrap as `StartupException` or similar?

imeszaros:
We can do it if it's not a problem that it changes the existing behavior. But then for the sake of consistency we should also wrap the exception rethrown in `Jooby.start()`, what do you think?

jknack:
I remembered someone to ask for this in 1.x. If you think is a good, let's add it. Otherwise, looks perfect.

imeszaros:
I've done it both places.

jooby/src/main/java/io/jooby/Jooby.java
```diff
@@ -1040,7 +1041,16 @@ public static Jooby createApp(@Nonnull String[] args, @Nonnull ExecutionMode exe
...
 
-    Jooby app = provider.get();
+    Jooby app;
+    try {
+      app = provider.get();
+    } catch (Throwable t) {
+      LoggerFactory.getLogger(Jooby.class)
+          .error("Application initialization resulted in exception", t);
+
+      throw new StartupException("Application initialization resulted in exception", t);
```
jknack, line 1051: we shouldn't wrap if it is a startup exception already

PR #2136, opened 2020-11-23
Add Jasypt module for encrypted properties

### LINKED ISSUE #2127: jasypt: encrypt properties
```
{
   install(new JasyptModule(key));
}
```

application.conf:

```
jasypt.sensible.prop = encrypted
```

Module will process all the keys prefixed with `jasypt`, decrypt them and rewrite as `sensible.prop`

jknack:
- Added Environment.setConfig(Config) too

Fixes #2127

### DIFF
--- a/modules/jooby-jasypt/src/main/java/io/jooby/jasypt/JasyptModule.java
+++ b/modules/jooby-jasypt/src/main/java/io/jooby/jasypt/JasyptModule.java
@@ -1,0 +1,183 @@
+/**
+ * Jooby https://jooby.io
+ * Apache License Version 2.0 https://jooby.io/LICENSE.txt
+ * Copyright 2014 Edgar Espina
+ */
+package io.jooby.jasypt;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Consumer;
+
+import javax.annotation.Nonnull;
+
+import org.jasypt.encryption.pbe.PBEStringEncryptor;
+import org.jasypt.encryption.pbe.PooledPBEStringEncryptor;
+import org.jasypt.encryption.pbe.StandardPBEStringEncryptor;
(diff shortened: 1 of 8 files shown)


modules/jooby-jasypt/src/main/java/io/jooby/jasypt/JasyptModule.java
```diff
@@ -0,0 +1,183 @@
...
+    ServiceRegistry services = application.getServices();
+    services.put(PBEStringEncryptor.class, encryptor);
+    Class encryptorType = encryptor.getClass();
+    services.putIfAbsent(encryptorType, encryptor);
+  }
+
+  /**
+   * Creates a PBE encryptor from application property files. The given key is used for reading
+   * jasypt password and possible other customization. Please note the <code>key.password</code>
+   * must be present in other to configure proper password setup.
```
imeszaros, line 123: typo: 'in order' instead of 'in other'

modules/jooby-jasypt/src/main/java/io/jooby/jasypt/JasyptModule.java
```diff
@@ -0,0 +1,183 @@
...
+   *
+   * Properties properties:
```
imeszaros, line 125: typo: 'Properties' is duplicated

imeszaros (approved):
Cool! For me, a total newbie with Jasypt, only one thing is unclear, how do I encrypt my passwords in the first place? But I guess who wants to use this module knows it already :D

jknack:
Good question. I do it manually... using `encryptor.encrypt()`. Isn't that hard that is way I didn't add anything.

imeszaros:
Ah i found they have a command line utility for that: http://www.jasypt.org/cli.html

jknack:
That's cool! I didn't know it, we could probably add a note/link to the doc.

PR #2171, opened 2020-12-08
netty: Fix memory leak on HTTP post

### LINKED ISSUE #2167: Memory leak in the jooby-netty implementation
In our project we use Jooby, and have more recently upgraded to version 2, currently running on the latest version 2.9.4.

I've spent a couple of days hunting down a memory leak within our application and have found out that the class:
`io.netty.handler.codec.DefaultHeaders$HeaderEntry` is causing some issue.

**How to recreate:**
Create an empty project with latest version of Jooby + Netty and Gradle. Run a POST-request to the running server, examine that the above mentioned class increases for every call (which is normal), but GC only manages to handle about 60% of it.

I have also created a project with Ktor + Netty, but this does NOT happen there, GC manages to clear what it should.

I have not looked into the implementation here, but hopefully you might now where to look?
It does seem to only apply to POST-requests, and it is quite hard to see it, we found it because we did not deploy for 4-5 days straight.

We have swapped to Jetty to test, and cannot see any memory leakage like this.

Please let me know if more information is required to solve this issue

jknack:
- Clean up thread local request references used by DefaultHttpDataFactory
- This was causing some io.netty.handler.codec.DefaultHeaders$HeaderEntry to leak
- Fix works, but not sure how to write a test to demonstrate

Fixes #2167

### DIFF
--- a/modules/jooby-netty/src/main/java/io/jooby/internal/netty/HttpRawPostRequestDecoder.java
+++ b/modules/jooby-netty/src/main/java/io/jooby/internal/netty/HttpRawPostRequestDecoder.java
@@ -6,8 +6,10 @@
 package io.jooby.internal.netty;
 
 import io.netty.handler.codec.http.HttpContent;
+import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.LastHttpContent;
 import io.netty.handler.codec.http.multipart.HttpData;
+import io.netty.handler.codec.http.multipart.HttpDataFactory;
 import io.netty.handler.codec.http.multipart.HttpPostRequestDecoder;
 import io.netty.handler.codec.http.multipart.InterfaceHttpData;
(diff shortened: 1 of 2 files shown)


imeszaros (approved):
I wonder how Undertow will be after the migration to Netty... hopefully then Jooby can drop the complicated server impl. for Netty. :D

jknack:
Probably yea... I think is is bad decision

hc-codersatlas:
Would rather undertow be dropped when that happens. Not enough support from IBM going forward. Only a single maintainer (mostly).  No longer a core focus (moved to Quarkus / Vertx / Netty).

Undertow is/has been in maintenance mode for a long time. Meanwhile Netty is adding quic and io_uring, which are both quite interesting. It's a bit sad - Undertow was the replacement for Netty back in the days.

PR #2203, opened 2020-12-20
HibernateModule : Accept list in constructor and scan

hc-codersatlas:
HibernateModule - accept list for constructor and scan to avoid copying list to a new array.


jknack:
yea, we can add it but from usage perspective vararg is better:

```
new HibernateModule(Bean1.class, Bean2.class)
```

hc-codersatlas:
Use case is scanning for list of classes, which returns a list. To use vararg the list needs to be converted into an array which then converts back into a list hence this PR :).

PR #2327, opened 2021-04-27
fixes #2324 getRequestURL(path) should not include current request's query string

### LINKED ISSUE #2324: getRequestURL(path) adds irrelevant queryString from current request
This line probably should be removed:
https://github.com/jooby-project/jooby/blob/34eb70cf2583e317d3c77f94d165bc8be7520bb5/jooby/src/main/java/io/jooby/DefaultContext.java#L252

If getRequestURL() is used with a different path than current request, then adding query string from current request is very unexpected and results in bugs, e.g. when implementing an OAuth2 client. Or this line could be moved to the empty path if branch, e.g. if user wants to get the current request's full URL.

cbxp:


jooby/src/main/java/io/jooby/DefaultContext.java
```diff
@@ -228,10 +228,10 @@
   }
 
   @Override default @Nonnull String getRequestURL() {
-    return getRequestURL("");
+    return getRequestURL(null);
   }
 
-  @Override default @Nonnull String getRequestURL(@Nonnull String path) {
```
jknack, line 234: let's keep the `@Nonnull` annotation. This always append the provided path (never the `getRequestPath` and never append the queryString

jooby/src/main/java/io/jooby/DefaultContext.java
```diff
@@ -228,10 +228,10 @@
   }
 
   @Override default @Nonnull String getRequestURL() {
-    return getRequestURL("");
+    return getRequestURL(null);
```
jknack, line 231: This one add call `getRequestURL(getRequestPath())` and append queryString if present

jknack:
Looks good, couple of comment. Let's make sure javadoc is updated too

jooby/src/main/java/io/jooby/DefaultContext.java
cbxp: The annotation is still in the interface (for external consumers), but locally, there was a null check already, so why not pass a null as "undefined" value? Empty string can actually have a meaning of empty path

jooby/src/main/java/io/jooby/DefaultContext.java
cbxp: OK, this is probably a better idea, yes

angryziber:
@jknack so, is it ok now to merge?

PR #2328, opened 2021-04-28
Flexible coroutine context

### LINKED ISSUE #2257: How to redefine CoroutineContext
When using coroutine router, it has a final coroutineScope inside, which is created lazily from worker.

However, there seems to be no easy way to add a CoroutineContext to the scope, e.g. MDCContext from 
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-slf4j/index.html

Could there be a setter for coroutineScope, to enable code like this:

```kotlin
coroutine {
  coroutineScope += MDCContext()

  // routes here
}
```

or another option is to add scope to CouroutineRouter constructor:

```kotlin
coroutine(context = MDCContext()) {
  // routes here
}
```

cbxp:
Make it possible to redefine coroutine context for every launch to be able to properly implement suspend/restore of additional state, e.g. MDC, transactions, etc

Fixes #2257

### DIFF
--- a/jooby/src/main/kotlin/io/jooby/CoroutineRouter.kt
+++ b/jooby/src/main/kotlin/io/jooby/CoroutineRouter.kt
@@ -5,16 +5,10 @@
  */
 package io.jooby
 
-import kotlinx.coroutines.CoroutineExceptionHandler
-import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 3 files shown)


jknack (changes requested):
Looks good, can you revert all the code format changes?

Thank you.

PR #2354, opened 2021-05-26
Jooby.install(Supplier<Jooby>) doesn't share registry fix #2240

### LINKED ISSUE #2240: Jooby.install(Supplier<Jooby>) doesn't share registry
Hey,

Thanks for the #1905 feature! It helps with app composition a lot. But I noticed a behavior I didn't expect: parent module doesn't share the registry with a child component.

```kotlin
class App : Kooby({
    install(GuiceModule())
    install(::ChildApp)

   get("/child") { require(Service::class).foo() } // works
})

class ChildApp : Kooby({
    get("/child") { require(Service::class).foo() } // throws RegistryException
})

class Service {
    fun foo() {}
}
```

I figured out how to make it work but it feels like a workaround:
```kotlin
class App : Kooby({
    install(GuiceModule())
    install(::ChildApp)

    onStarting {
        services.put(Service::class.java, require(SettlementRepository::class))
    }

    get("/child") { require(Service::class).foo() } // works
})

class ChildApp : Kooby({
    get("/child") { require(Service::class).foo() } // works
})

class Service {
    fun foo() {}
}
```

jknack:


imeszaros (approved):
Seems fine to me!

Issue #2357, opened 2021-05-27
Jooby header(String) doesn't support case insensitive


rlapao:
Although the documentation says so (https://www.javadoc.io/doc/io.jooby/jooby/2.9.1/io/jooby/Context.html#header-java.lang.String-) the method header when receiving for example a lowerCase name, when the request had an upperCase name, the method returns an empty value.

Hence, this fails to comply with the HTTP Headers RFC.

Tested in version 2.9.1

PR #2358, opened 2021-05-28
Fixes case insensitive headers

froque:
Fixes issue #2357


jooby/src/main/java/io/jooby/internal/HashValue.java
```diff
@@ -26,11 +26,11 @@
 import java.util.function.BiConsumer;
 
 public class HashValue implements ValueNode, Formdata {
```
jknack, line 28: What if we allow hashvalue to get a `Supplier<Map<String, ValueNode>>` at construction level, then we add a `Value.headers` static method and from there we provide a `TreeHashMap`?

jknack (changes requested):

jooby/src/main/java/io/jooby/internal/HashValue.java
froque: changed as requested

jknack (approved):

PR #2361, opened 2021-06-04
Add test scope to usages of mockito-core in jooby modules

froque:
The mockito dependency is not needed in production code and should not be
consumed by consuming applications.


jknack (commented):
Are these the only one without test scope?

PR #2362, opened 2021-06-09
Improve coroutine context redefinition introduced in Jooby 2.9.6

cbxp:
Reasons:

1. Extra coroutine context often needs to access Jooby Context
2. Relying on users to extend the default CoroutineContext can be error prone - users can forget to include the ExceptionHandler, resuling in "hanging" asynchronous requests

Signature of launchContext lambda has changed in not 100%-compatible way, but as launchContext was a recent new feature and wasn't yet documented, probably backwards-compatibility with previous API is not needed.

### DIFF
--- a/docs/asciidoc/responses.adoc
+++ b/docs/asciidoc/responses.adoc
@@ -625,7 +625,7 @@ Jooby allows you to use this experimental API by setting the `coroutineStart` op
 [source, kotlin]
 ----
 {
-  coroutine (CoroutineStart.UNDISPATCHED) {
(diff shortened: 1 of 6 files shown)


cbxp:
@jknack do you really want us to sign the commits?

angryziber:
@jknack how about this one?

cbxp:
@jknack it would be really nice to merge this one as I am talking about Jooby+coroutines at conferences and this issue is critical for the proper usage.

jknack:
yea, will try to review everything over the weekend. I'm busy and sick.

jknack (approved):

PR #2367, opened 2021-06-13
Use sendError instead of send for NOT_FOUND in AssetHandler

GitRowin:


jknack (commented):
This was intentional. In my experience assets (static resources) should not generate noise (that is why just send a `404`). Cached browser resources, missing map files, bot attacks cause a lot of noise.

GitRowin:
I use this asset handler: `app.assets("*", "static");`

Now I'm unable to change any 404 page. What do you mean by noise? If you're talking about the NotFoundException spam, the end user can just add a 404 error handler that doesn't call `cause.printStackTrace();`. However, I don't think a 404 should cause a stack trace to be printed in the first place. Jooby should just catch the exception and return the default 404 page.

jknack:
> What do you mean by noise? If you're talking about the NotFoundException spam,

Right.

> However, I don't think a 404 should cause a stack trace to be printed in the first place

Depends on business/application. For static resources doesn't make sense. That's why uses a 404 response.

The `AssetHandler` allows you to add your own `AssetSource`:

```java
AssetSource throwError = path -> throw new NotFoundException(path);

app.assets("*", new AssetHandler(AssetSource.create(Paths.get("static"), throwError));

```

So you add. all the asset source you want (file system, classpath, etc) and put your on the end that throws an exception.

Exception will be caught by error handler and then you can do what ever you want there.

GitRowin:
Clever solution, that works for me. What about logging all requests, similar to Nginx's and Apache's `access.log`? `after` handlers don't get called for 404 requests.

Also, when a handler throws an exception and Jooby returns status code 500, `ctx.getResponseCode()` returns `OK` in `after` handlers.

jknack:
> Also, when a handler throws an exception and Jooby returns status code 500, ctx.getResponseCode() returns OK in after handlers.

Pushed a fix: https://github.com/jooby-project/jooby/commit/c7d1bccf130f618bae0c58aff9070a77cdee6fde

> What about logging all requests, similar to Nginx's and Apache's access.log? after handlers don't get called for 404 requests.

There is an [AccessLogHandler](https://jooby.io/#handlers-accessloghandler). Added an example on commit too:

    [25/Jul/2021:12:08:50 -0300] "GET /2367/index.js HTTP/1.1" 404 0 13

PR #2368, opened 2021-06-14
Adds support for JSON-B (JSR 367)

froque:
Adds yasson dependecy as the implementation for JSON Binding.

There was a previous similar addition for jooby 1 that was not included in jooby 2:
https://github.com/jooby-project/jooby/pull/1248
https://jooby.io/v1/doc/yasson/

### DIFF
--- a/modules/jooby-jsonb/pom.xml
+++ b/modules/jooby-jsonb/pom.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <parent>
+    <groupId>io.jooby</groupId>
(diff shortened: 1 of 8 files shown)


modules/jooby-jsonb/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+
+    <dependency>
+      <groupId>jakarta.json.bind</groupId>
+      <artifactId>jakarta.json.bind-api</artifactId>
+      <scope>compile</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.eclipse</groupId>
+      <artifactId>yasson</artifactId>
```
froque, line 30: It is not clear to me if jooby should provide the jsonb implementation or let the client choose the implementation.

jknack: yea, we normally use/follow the name of the library: `yasson` not the specification `jsonb`, so it should be `joooby-yasson`

froque: @jknack I don't agree with that. If a user wants to use another implementation JSON-B implementation as https://johnzon.apache.org/ it will still be required to use a Yasson module. 

Anyway I implemented as requested in another pull request #2383

jknack (approved):

PR #2402, opened 2021-07-27
Implemented getting client certificates from the context for all 3 se…

### LINKED ISSUE #2401: Is it possible to get peer certificate from ctx when clientAuth == REQUIRED?
Hi,

   I have been trying to get the certificate presented by the client from the request context, and I haven't found a way to do it. Mutual TLS works, so I know the server has the peer's certificate. It seems that all access to Netty's SSLEngine is blocked, so I can't pull it directly from Netty, and the context doesn't have any notion of channels, so I can't get it from there either.  I'm hoping this is just a difficult thing to find, and that it is possible to get the peer certificates from the request. (Other frameworks handle this use case, so this isn't a unique concern)

    Please let me know if I've misunderstood, and there is a way to get to the peer certificate/sslEngine.

cmitchell:
Should resolve #2401.  

Added a "getClientCertificates()" function to the context that returns the certificates presented by the client during the SSL handshake, and implemented for all 3 servers.  I didn't see any tests that setup/verified client authentication certificates, or even used ClientAuth.REQUIRED, therefore I didn't modify or add any tests. I have verified that this works for netty (and probably also utow since it's practically the same code) through actual code.

### DIFF
--- a/jooby/src/main/java/io/jooby/Context.java
+++ b/jooby/src/main/java/io/jooby/Context.java
@@ -21,6 +21,7 @@ import java.nio.channels.FileChannel;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
+import java.security.cert.Certificate;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
@@ -626,6 +627,13 @@ public interface Context extends Registry {
    * @return The name of the protocol the request. Always in lower-case.
    */
   @Nonnull String getProtocol();
+
+  /**
+   * The certificates presented by the client for mutual TLS. Empty if ssl is not enabled, or client authentication is not required.
+   *
+   * @return The certificates presented by the client for mutual TLS. Empty if ssl is not enabled, or client authentication is not required.
+   */
+  @Nonnull Certificate[] getClientCertificates();
 
   /**
    * Server port for current request.
--- a/jooby/src/main/java/io/jooby/Context.java
+++ b/jooby/src/main/java/io/jooby/Context.java
@@ -633,7 +633,7 @@ public interface Context extends Registry {
    *
    * @return The certificates presented by the client for mutual TLS. Empty if ssl is not enabled, or client authentication is not required.
    */
-  @Nonnull Certificate[] getClientCertificates();
+  @Nonnull List<Certificate> getClientCertificates();
 
   /**
    * Server port for current request.
--- a/modules/jooby-test/src/main/java/io/jooby/MockContext.java
+++ b/modules/jooby-test/src/main/java/io/jooby/MockContext.java
@@ -541,8 +541,8 @@ public class MockContext implements DefaultContext {
     return "HTTP/1.1";
   }
 
-  @Nonnull @Override public Certificate[] getClientCertificates() {
-    return new Certificate[0];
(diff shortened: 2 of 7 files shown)


jooby/src/main/java/io/jooby/Context.java
```diff
@@ -627,6 +628,13 @@
    */
   @Nonnull String getProtocol();
 
+  /**
+   * The certificates presented by the client for mutual TLS. Empty if ssl is not enabled, or client authentication is not required.
+   *
+   * @return The certificates presented by the client for mutual TLS. Empty if ssl is not enabled, or client authentication is not required.
+   */
+  @Nonnull Certificate[] getClientCertificates();
```
jknack, line 636: I do prefer `List<Certificate>` over `Certificate[]`

modules/jooby-test/src/main/java/io/jooby/MockContext.java
```diff
@@ -540,6 +541,10 @@ public MockContext setResponseType(@Nonnull MediaType contentType, @Nullable Cha
     return "HTTP/1.1";
   }
 
+  @Nonnull @Override public Certificate[] getClientCertificates() {
```
jknack, line 544: let's add an attribute and a setter

jknack (commented):
Love it! Very clean. I need to figure it out how to add a test with client certificate

jooby/src/main/java/io/jooby/Context.java
cmitchell: The underlying method that gets called is getPeerCertificates() and it returns an array. (https://docs.oracle.com/en/java/javase/11/docs/api/java.base/javax/net/ssl/SSLSession.html#getPeerCertificates() for netty
https://undertow.io/javadoc/2.1.x/io/undertow/server/SSLSessionInfo.html#getPeerCertificates--  for undertow
For jetty, you already add the `SecureRequestCustomizer())`, which adds that javax.servlet.request.X509Certificate as the key, and puts the value in the map as an array)
The order of the certificates in the array matters. From the java documentation (first link) 
```
Returns:
an ordered array of peer certificates, with the peer's own certificate first followed by any certificate authorities.
```
If you return a List, it could not be a generic List. It would have to be an implementation of List that maintains order. Making this change would introduce the risk of messing up the order during the transfer from array to the list, and most callers would expect it to be an array.  Your decision. 


Side note, I do see from https://www.eclipse.org/jetty/javadoc/jetty-11/constant-values.html#org.eclipse.jetty.server.SecureRequestCustomizer.JAKARTA_SERVLET_REQUEST_X_509_CERTIFICATE that as of Jetty 11, it should be `jakarta.servlet.request.X509Certificate`. I can just use the static field.  I can make this update when you decide whether this return type should remain an array, or change to a list.

modules/jooby-test/src/main/java/io/jooby/MockContext.java
cmitchell: Will do

jooby/src/main/java/io/jooby/Context.java
jknack: List in general are better than array. A call to: `Arrays.asList()` will keep the order.

No jetty 11 for now. I'm planning to make 3.x to require at least Java 11 and definitely upgrade jetty to 11 or whatever.

Issue #1805, opened 2020-06-21
Openapi generation fails if I use java.net.URI as a query param


extbe:
Hi!

I try to use `java.net.URI` as a query parameter in MVC controller
```java
@Path("/")
public class Controller {

  @GET
  public String sayHi(@QueryParam URI uriParam) {
    return "Welcome to Jooby!";
  }
}
``` 

And `jooby:openapi` fails with error:
`
[ERROR] Failed to execute goal io.jooby:jooby-maven-plugin:2.8.7:openapi (default-cli) on project test-app: execution of openapi resulted in exception: Null key for a Map not allowed in JSON (use a converting NullKeySerializer?) (through reference chain: io.jooby.internal.openapi.OpenAPIExt["components"]->io.swagger.v3.oas.models.Components["schemas"]->java.util.LinkedHashMap["null"]) -> [Help 1]
`

The same error occurs when I use `java.net.URL`

PR #2448, opened 2021-10-27
Fix OpenAPI generation for newer Kotlin compiler and support java.time types

cbxp:
This fixes the latest Jooby 2.11.0:

1. It seems that newer Kotlin compiler (e.g. 1.5.30, with IR backend) sometimes names the Continuation parameter as `$completion` in suspending functions. Let's add support for it, or otherwise OpenAPI generation crashes with messages that parameter type cannot be found for `$completion`.
2. Not related to Kotlin: do not crash when routes have LocalDate, LocalDateTime, Instant, etc parameters - use the correct DateSchema or DateTimeSchema for these.

Having these fixes, we now can successfully generate openapi spec for our project. Without these fixes, openapi crashes with mostly meaningless error messages, like in issue #1805

### DIFF
--- a/jooby/src/main/kotlin/io/jooby/Kooby.kt
+++ b/jooby/src/main/kotlin/io/jooby/Kooby.kt
@@ -133,7 +133,7 @@ open class Kooby constructor() : Jooby() {
   }
 
   @RouterDsl
-  fun after(handler: AfterContext.() -> Any): Kooby {
+  fun after(handler: AfterContext.() -> Unit): Kooby {
     super.after { ctx, result, failure -> AfterContext(ctx, result, failure).handler() }
     return this
   }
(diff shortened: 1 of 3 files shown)


cbxp:
@jknack any chance to merge the fix or do we have to fork and release our own jooby? ;-)

jknack:
didn't ;look yet. Will try to release next weekend. Don't fork be patient 😄

jknack (approved):

sillen102:
I'm having this issue. If I use LocalDateTime and/or ZoneId as @QueryParam I get:
`Null key for a Map not allowed in JSON (use a converting NullKeySerializer?) (through reference chain: io.jooby.internal.openapi.OpenAPIExt["components"]->io.swagger.v3.oas.models.Components["schemas"]->java.util.LinkedHashMap["null"])`

Issue #2486, opened 2021-12-22
jooby-apt: Module class not generated when base class is an external class (from another project/jar)


JozefDropco:
Hello, 
I am a newbie in using Jooby. We are about to migrate all our applications from JavaEE to Jooby and Undertow. I am trying to do it for one of our simplest application. I was able to setup PAC4J for security, servlets to run them as we were used to before (we will decommission them later), get running most of our webservices with JAX-RS. But one of them is not working. We have a common library were we have OperationWebService which is used for health check and also some insight to the application settings. This service is then extended in each application.
**Common from library**
```java
public class OperationWebService {

    @GET
    @Path("/clue")
    @Produces({MediaType.APPLICATION_JSON, MediaType.TEXT_PLAIN})
    public final Response getClue(final @Context HttpHeaders headers) {
        //...
    }

    @GET
    @Path("/config")
    @Produces({MediaType.APPLICATION_JSON, MediaType.TEXT_PLAIN})
    public final Response getConfig(final @Context HttpHeaders headers) {
      // ...
    }

    @GET
    @Path("/state")
    @Produces({MediaType.APPLICATION_JSON, MediaType.TEXT_PLAIN})
    public final Response getState(final @Context HttpHeaders headers) {
        // ....
    }

    @GET
    @Path("/introspection")
    public final Response getIntrospection(final @Context HttpHeaders headers) {
     //.....
    }

    @GET
    @Path("/idle")
    @Produces({MediaType.APPLICATION_JSON, MediaType.TEXT_PLAIN})
    public final Response peekIdle() {
     //  .......
    }
```

Each application then extends this OperationWebService and can override few methods. But the JAX-RS annotated methods are final and cannot be overriden.

```java
@Path("/ws/ops")
public class FiestaOpsWebService extends OperationWebService {
}
```

Am I missing some settings for maven-compiler-plugin?
```xml
            <plugin>
                <artifactId>maven-compiler-plugin</artifactId>
                <version>3.8.1</version>
                <configuration>
                    <source>1.8</source>
                    <target>1.8</target>
                    <annotationProcessorPaths>
                        <path>
                            <groupId>io.jooby</groupId>
                            <artifactId>jooby-apt</artifactId>
                            <version>${jooby.version}</version>
                        </path>
                    </annotationProcessorPaths>
                    <compilerArgs>
                        <compilerArg>
                            -Ajooby.debug=true
                        </compilerArg>
                        <compilerArg>
                            -Ajooby.services=true
[...]

PR #2492, opened 2022-01-04
https://github.com/jooby-project/jooby/issues/2486

JozefDropco:
Hi, I found the issue in jooby-apt and added the support there for this case. I have disabled it by default to decrease the impact on other community.


modules/jooby-apt/src/main/java/io/jooby/apt/JoobyProcessor.java
```diff
@@ -51,26 +36,28 @@
...
-   *   HTTP_METHOD: [method1, ..., methodN]
+   * HTTP_METHOD: [method1, ..., methodN]
    * }
    */
   private Map<TypeElement, Map<TypeElement, List<ExecutableElement>>> routeMap = new LinkedHashMap<>();
 
   private boolean debug;
   private boolean incremental;
   private boolean services;
+  private boolean inspectSuperTypes;
```
jknack, line 55: we need to find another name to better the describe it. Inherited does work as long as classes are on same project (as you already figured it out).

JozefDropco: HI @jknack I have renamed it to inspectSubClasses as its really what it does

jknack: Still found a bit confusing, bc inherited works without this option enabled. The issue raises when class isn't part of the project (source code)

PR #2512, opened 2022-01-26
Fix Map and byte[] for openapi

Xerxekyran:
This pull requests contains two fixes when generating the openapi defintion using jooby-openapi:

**Fix for Map:**
According to https://swagger.io/docs/specification/data-models/dictionaries/ a map should be represented by a schema that has the type as well as an additionalProperties element with the type property set to the value property of the map.

So the following map

```
{
  "en": "English",
  "fr": "French"
}
```

should become the following schema:

```
type: object
  additionalProperties:
    type: string
```

**Fix for byte[]:**
There exists a special ByteArraySchema class that can be used in this case, as it produces the correct schema for it:

```
schema": {
  "type": "string",
  "format": "byte"
}
```


Xerxekyran:
Shall I rebase the branch or fix the checkstyle problem that was introduced two days ago?

jknack:
Fixed it. Thanks.

Build is running again

PR #2632, opened 2022-09-16
3.0x Slf4j 2.0 upgrade

agentgt:


jknack:
will manually apply the sl4fj changes due this PR has lot more of commits

agentgt:
I can squash it if you like if that makes it easier. Ditto for the java modules branch.

PR #2667, opened 2022-11-02
Added jax rs path param

dmitryb-dev:


jknack:
Thank you. Can you send it against 3.x branch?

Issue #2864, opened 2023-04-04
[FR] JStachio module


agentgt:
Hey @jknack I have a simple PR I am working on that adds my templating engine JStachio to 3.x `jooby/modules`. 

https://github.com/jstachio/jstachio

Doc: https://jstach.io/jstachio/

JStachio is basically a mustache implementation that compiles to Java code.

The PR will be similar to the Rocker implementation as JStachio is type-safe and compiled. Currently JStachio outperforms Rocker but is negligible slower than JTE which is pretty good IMO because I consider the other two basically JSP (aka Java code pretending to be a templating). 

While I love handlebars.java for user generate templates I/we have been using my library for server side templates as well as for code generation. I would love your input on the library as I have used Handlebars.java for many years. I plan on making it 1.0.0 this month.

As you might recall I have been sort of against modules :) so it is ironic making this feature request but my hope is to perhaps to replace the techempower rocker `fortunes.html` with my library as well as offer a fast type safe templating language to Jooby.

If Jooby is too late in the 3.x RCing and/or you do not want the distraction I can try to make it a module on JStachio itself.

jknack:
> As you might recall I have been sort of against modules :) so it is ironic making this feature request but my hope is to perhaps to replace the techempower rocker fortunes.html with my library as well as offer a fast type safe templating language to Jooby.

Indeed 😄 

Happy, to adding your JStachio as jooby module

agentgt:
@jknack I did some perf testing with TechEmpower benchmark and replaced Rocker with JStachio without any buffering... just raw new StringBuilder on each request and it appears to perform roughly 25% faster on the Fortunes test than rocker (which is the only test that I know of that uses templating).

I have to finish making the test truly apples to apples as Rocker had some issues with JDK 17 which JStachio requires. This is also jooby 2.9. 

I thought for sure I was going to have to do some byte caching and or threadlocal magic but it appears that it doesn't matter at least on my machine.
My guess is the techniques that Rocker did with byte arrays probably matter far less with modern JVMs.

I would imagine JTE (rockers major competitor) would perform similarly better albeit I'm hoping we go with JStachio given the example by [TechEmpower is a mustache template](https://github.com/TechEmpower/FrameworkBenchmarks/wiki/Project-Information-Framework-Tests-Overview#minimum-template).

[...]

PR #2865, opened 2023-04-05
JStachio module #2864

agentgt:
Add jstachio support to jooby.

### DIFF
--- a/modules/jooby-jstachio/src/main/java/io/jooby/jstachio/JStachioMessageEncoder.java
+++ b/modules/jooby-jstachio/src/main/java/io/jooby/jstachio/JStachioMessageEncoder.java
@@ -1,0 +1,72 @@
+/*
+ * Jooby https://jooby.io
+ * Apache License Version 2.0 https://jooby.io/LICENSE.txt
+ * Copyright 2023 Edgar Espina
+ */
+package io.jooby.jstachio;
+
+import java.nio.charset.Charset;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.jooby.Context;
+import io.jooby.MessageEncoder;
+import io.jstach.jstachio.JStachio;
+
+/*
+ * Should this be public?
+ */
+class JStachioMessageEncoder implements MessageEncoder {
+
+  private final JStachio jstachio;
+  private final Charset charset;
+  private final int bufferSize;
+
+  public JStachioMessageEncoder(@NonNull JStachio jstachio, @NonNull Charset charset, int bufferSize) {
+    super();
+    this.jstachio = jstachio;
+    this.charset = charset;
+    this.bufferSize = bufferSize;
+  }
+
+  @Override
+  public byte[] encode(Context ctx, Object value) throws Exception {
+    if (supportsType(value.getClass())) {
+      return render(value);
+    }
+    return null;
+  }
+  
+  protected boolean supportsType(Class<?> modelClass) {
+    return jstachio.supportsType(modelClass);
+  }
+
+  protected byte[] render(Object value) {
+    StringBuilder b = aquireBuffer();
+    try {
+      jstachio.execute(value, b);
+      return b.toString().getBytes(charset);
+    } finally {
+      releaseBuffer(b);
+    }
+  }
+
+  /**
+   * Returns a new buffer. {@link #releaseBuffer(StringBuilder)) should be
+   * called when done.
+   * 
+   * @return a buffer either from a pool or a new one.
+   */
+  protected StringBuilder aquireBuffer() {
+    return new StringBuilder(bufferSize);
+  }
+
+  /**
+   * Releases the buffer. Override for pooling buffers.
+   * 
+   * @param sb
+   */
+  protected void releaseBuffer(StringBuilder sb) {
+    return;
+  }
+
+}
--- a/pom.xml
+++ b/pom.xml
@@ -352,7 +352,7 @@
       </dependency>
 
       <dependency>
-        <groupId>io.jstach</groupId>
+        <groupId>io.jooby</groupId>
         <artifactId>jooby-jstachio</artifactId>
         <version>${jooby.version}</version>
       </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -21,6 +21,7 @@
     <!-- Parser and Renderer -->
     <freemarker.version>2.3.32</freemarker.version>
     <handlebars.version>4.3.1</handlebars.version>
+    <jstachio.version>0.12.0</jstachio.version>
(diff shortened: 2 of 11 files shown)


jknack:
@agentgt will review soon. In JStachio what did you do with buffer size? can you point me there?

I would like to add a event-loop buffer on jooby with a max of 8k (like you did) and reuse that buffer...

modules/jooby-jstachio/src/main/java/io/jooby/jstachio/JStachioMessageEncoder.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+  }
+
+  /**
+   * Returns a new buffer. {@link #releaseBuffer(StringBuilder)) should be
+   * called when done.
+   * 
+   * @return a buffer either from a pool or a new one.
+   */
+  protected StringBuilder aquireBuffer() {
```
agentgt, line 59: @jknack Per your question on buffering and potentially pooling. I use StringBuilder and no pooling at the moment. There has been lots of debate whether or not to do the whole threadlocal pooling given loom and how good newer JDKs are object creation.

agentgt:
I added a comment but have copied it here:

I use StringBuilder and no pooling at the moment. A new string builder initialized with a bufferSize is created everytime. 

There has been lots of debate whether or not to do the whole threadlocal pooling given loom and how good newer JDKs are object creation.

Even JTE the closest in terms of performance to JStachio removed their pooling (both JTE and mine are far faster (~ 1.7x) than Rocker):

https://github.com/casid/jte/releases/tag/2.2.0

That being said I added pretend pooling methods. Also I wasn't sure if MessageEncoder was going to change to ByteBuffer or something similar.

I guess I could try to load up whatever sandbox techempower has to test Rocker's pooling my naive StringBuilder.

Pooling is obviously going to win in isolated perf environments like JMH but from my testing using the template engine benchmark: https://github.com/agentgt/template-benchmark/ it only makes like a 10% difference on JDk 17.

jknack:
will look into JTE. I got and agreed on pooling+threadlocal.... the thing is: almost all the tests in tech-empower they make use and abuse of threadlocal and without them looks like jooby is not as fast, but that isn't truth.

something similar happens with router implementations... most of them are just switch/case statements :S

agentgt:
I can certainly add some implementation like I did for the template benchmarks: https://github.com/agentgt/template-benchmark/blob/ebaee9d71cccc41aba65c07b16ac76e1b2d29bd4/src/main/java/com/mitchellbosecke/benchmark/JStachio.java#L19

Which btw to your point on benchmarks... most forks of that template-benchmark do that little threadlocal cheat to their own library but fail to apply it to the others. That and they enable caching (compiled) but not for the others. It's one of the reasons Handlebars does not look as good on those benchmarks.

What might be better if we really need pooling is some sort of better `MessageEncoder` interface where you acquire a buffer and then release it and then I guess have that tied into the event loop buffers.

I guess the other issue w/ buffering and `MessageEncoder` is good ole team reactive. JStachio is not reactive and neither is `MessageEncoder` but team reactive requires everything be "reactive" (buffering is a form of backpressure solution but not good enough for them). I would understand the need for JStachio to be reactive if Java was lowlevel like Rust where you have intricate control of memory but most people fill up a model with plain collections (e.g. `List` instead of `Flow.Publisher`) and then render. Anyway that is my rant for the day.

agentgt:
@jknack let me know if you want me to squash the PR or any other changes. 

Right now the version is 0.12.0 but sometime next week it should be 1.0.0.

jknack:
you need to add .adoc file

agentgt:
Hopefully the doc is OK. It has been a long time since I have used adoc.

I'm not entirely sure how I can build a preview.

agentgt:
Hmm I get an artifact missing error:

```
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:3.1.0:java (asciidoctor) on project docs: An exception occurred while executing the Java class. asciidoctor: FAILED: /Users/agent/projects/jooby/docs/asciidoc/modules/jstachio.adoc: Failed to load AsciiDoc document: >>usage: Missing artifact: jooby-jstachio -> [Help 1]
```


I'm guessing I forgot to add jooby-jstachio to the bom or root pom (assuming adoc can do maven dependencies or that is a nifty plugin you wrote)?

agentgt:
Alright so now I get this error message:

```
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:3.1.0:java (asciidoctor) on project docs: An exception occurred while executing the Java class. File: /Users/agent/projects/jooby/docs/asciidoc/site/modules/jstachio/index.html: Cannot invoke "org.jsoup.nodes.Element.text()" because the return value of "org.jsoup.nodes.Element.selectFirst(String)" is null -> [Help 1]
```

Which I assume is some post processing you are running and missed putting something that the jsoup is expecting.

pom.xml
```diff
@@ -350,6 +351,12 @@
         <version>${jooby.version}</version>
       </dependency>
 
+      <dependency>
+        <groupId>io.jstach</groupId>
```
jknack, line 355: io.jooby

agentgt: I probably should have opened intellij or eclipse instead of using vim :)

agentgt:
```
Caused by: java.lang.NullPointerException: Cannot invoke "org.jsoup.nodes.Element.text()" because the return value of "org.jsoup.nodes.Element.selectFirst(String)" is null
    at io.jooby.adoc.DocGenerator.languageTab (DocGenerator.java:347)
    at io.jooby.adoc.DocGenerator.document (DocGenerator.java:320)
    at io.jooby.adoc.DocGenerator.lambda$generate$6 (DocGenerator.java:122)
```

Ok I see I messed something up on the language tab.

agentgt:
That `{somedepVersion}` is nice. I had such a pain in the ass getting Javadoc to do similar and ended up doing what you did with a post processor right on the generated HTML itself.

agentgt:
@jknack Ok I think everything is working. I got the doc build to work. 

I'm not sure if JStachio currently will work with Kotlin (kapt) but I don't think that will change anything in this PR. 

Let me know if there is anything else you want me to change and/or squash.

Thanks for the help!

jknack:
the doc gen only expect a secondary language when `role="primary"` is present. If you removed that, then no kotlin (secondary thing) is required.

> That {somedepVersion} is nice. I had such a pain in the ass getting Javadoc to do similar and ended up doing what you did with a post processor right on the generated HTML itself.

doing doc is as hard as everything else. doc in 1.x was messy messy... This one is bit better

PR #2921, opened 2023-05-16
Add Rocker style pre-encoding for JStachio

agentgt:
@jknack I have updated the JStachio module so that it can take advantage of pre-encoding (the static parts of the template are generated as bytes[] in advance).

The reason I did this is that in the techempower benchmarks Rocker performed better than JStachio's naive `String.getBytes()`.

I also have added threadlocal buffering and removed setting charset.

Setting the charset was a bad idea anyway as JStachio unlike others cannot always re-read the template from the filesystem furthermore escaping is naturally tied to charset.

I noticed that Rocker just assumes MediaType.html (and thus UTF-8) so I have done likewise.

I just released jstachio 0.17.0 but it will take an hour or so before central shows it.

### DIFF
--- a/modules/jooby-jstachio/src/main/java/io/jooby/jstachio/ByteBufferedOutputStream.java
+++ b/modules/jooby-jstachio/src/main/java/io/jooby/jstachio/ByteBufferedOutputStream.java
@@ -1,0 +1,117 @@
+/*
+ * Jooby https://jooby.io
+ * Apache License Version 2.0 https://jooby.io/LICENSE.txt
+ * Copyright 2014 Edgar Espina
+ */
+package io.jooby.jstachio;
+
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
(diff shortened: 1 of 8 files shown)


agentgt:
> @agentgt did you see the Guice release?
> 
> We should pick 7.0.0 right?

I assume you mean the DI framework. Yes I think so.

agentgt:
@jknack 

Here are the benchmarking results for JStachio to hopefully prove it as an OK replacement for Rocker in the TechEmpower benchmarks.

https://github.com/agentgt/template-benchmark/blob/utf8/README.md

I also ran the techempower benchmarks with older Jooby and backport of JStachio module and got similar results albeit the difference not as noticeable.

jknack:
@agentgt you need to rebase

agentgt:
@jknack I will tomorrow morning. I just noticed the conflict. Sorry about that.

Issue #3059, opened 2023-08-04
open-api generation doesn't work with indirection in jooby 3.0


tpoll:
In the process of upgrading from jooby 2.16.1 to 3.0.1, I noticed an issue with the open-api maven plugin generation.

While the plugin does run, it ends up not able to find any paths and produces a file like:

```json
{
  "openapi" : "3.0.1",
  "info" : {
    "title" : "Api1 API",
    "description" : "Api1 API description",
    "version" : "1.0"
  },
  "paths" : { }
}
```

I have a reproducible example here https://github.com/tpoll/jooby-example. You can replicate by running `mvn clean package`.

I believe the issue is related to the extra level of indirection I have via the `runner-core` module which accepts resources and then binds them directly.

If I remove the level of indirection in this project, the open-api spec seems to run fine.

jknack:
are you sure this same project work in 2.x?

Open API has no way of finding resources when bootstrap looks like:
```
public class Api1 {
  public static void main(String[] args) {
    JoobyRunner.create(args)
      .bindResource(new Resource1())
      .bindResource(new Resource2())
      .run();
  }
}
```

tpoll:
Yeah, I think it's actually taking advantage of the earlier issue I mentioned https://github.com/jooby-project/jooby/issues/2756 here. Since the old version seems to pick up on ever resource in the classpath, I believe it's still able to find the resources even with indirection. 

Is openapi specifically looking for the `mvc(new Controller()` calls to be in the same class where jooby itself is bootstrapped? I'm fine with adding some extra code on my side to make sure jooby can find the classes it needs to generate the openapi spec. 

I realize my use case is a little less common, but having a wrapper class around jooby is nice since I have multiple distinct APIs and we like to enforce common setup of various settings, filters, etc.

What do you think about adding an option into jooby-openapi to allow users to manually specify what resources they would like to be included. If that's something that sounds reasonable, I would be happy to try and work on a PR to add that functionality.

jknack:
yea, I'm not sure how 2.x could do it. The #2756 and their fix was a different isssue.

OpenAPI starts from the `applicationClass` you set as property in Maven/Gradle. From there it parses and scan all possible jooby idioms to find resources.

example::
```
runApp(args, App::new) => will scan from App
```

```
runApp(args, app -> {
  // will scan from lambda
});
```

It is the same for kotlin.

[...]

PR #3060, opened 2023-08-05
allow manual registration of mvc routes for openapi

tpoll:
@jknack  as discussed in https://github.com/jooby-project/jooby/issues/3059, this adds support to manually register a mvc class for open-api generation.

In my testing, order of the  `registerMvc` calls didn't break anything, just change the order of the generated paths in the `openapi.json` file.

**Open problem**:
Right now this doesn't actually scope the registration to a particular router.  Open to any ideas on how we could make that work. I thought about adding another param to `registerMvc`, but I didn't see an obvious way in code to match the value of that param to the owning routers in the parsing logic.

Very open to any feedback!


jknack:
we need a test that demonstrate how this works and helps (success case).

> Open problem: Right now this doesn't actually scope the registration to a particular router. Open to any ideas on how we could make that work. I thought about adding another param to registerMvc, but I didn't see an obvious way in code to match the value of that param to the owning routers in the parsing logic.

Another test that shows what the issue is.

tpoll:
Okay, I believe I have added tests that properly demonstrate the issue and how this works to resolve it.

I believe the `open problem` I earlier mentioned isn't actually an issue, since the openapi generator seems to always be scoped to a single jooby application class. I also added a test to demonstrate this.

jknack:
Thank you. It is clear now how it works. What if we add a custom annotation with clear documentation of when/how can be use it.

Something like:

```
@OpenAPIRegister({Resource1.class, Resource2.class})
public class Api1 {
    JoobyRunner.create(args)
      .bindResource(new Resource1())
      .bindResource(new Resource2())
      .run();
  }
}
```

> OpenAPIRegister
That or a better name.

So then the generator augment the resources found with what the annotation indicates.

tpoll:
I do like the annotation route, the latest commit I pushed switches over to that style. I'm not sure what the best name is either. I added the word `manual` to try and make it clear this annotation is only needed when one is trying to manually register.

I'm not sure where the annotation should live -- should it still live in the openapi module or should it live in jooby core?

Where would you like to see the documentation? On the annotation itself, or in the doc links for openapi?

modules/jooby-openapi/src/main/java/io/jooby/openapi/OpenApiManualRegister.java
```diff
@@ -0,0 +1,17 @@
...
+package io.jooby.openapi;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface OpenApiManualRegister {
```
jknack, line 15: add documentation here.

modules/jooby-openapi/src/test/java/issues/i3059/App3059A.java
```diff
@@ -0,0 +1,19 @@
...
+ * Jooby https://jooby.io
+ * Apache License Version 2.0 https://jooby.io/LICENSE.txt
+ * Copyright 2014 Edgar Espina
+ */
+package issues.i3059;
+
+import io.jooby.Jooby;
+import io.jooby.openapi.OpenApiManualRegister;
+
+@OpenApiManualRegister({ControllerA3059.class, ControllerB3059.class})
```
jknack, line 11: cool 😎

jknack (changes requested):

jknack:
> I'm not sure where the annotation should live -- should it still live in the openapi module or should it live in jooby core?

Let's add it on io.jooby.annotation package on jooby(-core)

tpoll:
I believe I have addressed the feedback, I also thought a bit more and moved back to the originally proposed name of `OpenAPIRegister`

jknack (approved):

PR #3076, opened 2023-08-12
router: non-ascii support

### LINKED ISSUE #3070: Static `assets` doesn't work with non-ASCII file names
When using the `assets` method to declare a static directory to serve files from, I can not access the files with the browser. I think this is because maybe the requests are not properly URL-decoded.

### LINKED ISSUE #3071: Can't create an endpoint that contains non-ASCII characters
I want to create some routes, that contain characters outside of the ASCII range. I think, this is a pretty reasonable request. All that needs to be done, is to, at compile time, encode the non-ASCII letters of endpoints (I am using the MVC style and the annotation processor). Otherwise, I need to write my endpoints with %ab%cd all over the place and it's not really nice. This could be an option as well. If my endpoints were not MVC, I would be able to just do the encoding myself, but for annotation, the values passed must be constant, so no concatenations nor encodings can be done.

I also considered adding some kind of handler or filter that would rewrite the path of my endpoints to be URI-encoded when they are added, but I am a new user of the Jooby library, and I have no idea if this is possible to do.

Example:
```java
public class Test {
    @GET(path = "/tést")
    public String test(Context ctx) { return "hello"; }
}
```

After adding this with `mvc(new Test())`, try accessing this endpoint. You will not be able to access it.

jknack:
- fix asset handler to work with non-ascii chars
- routes with non-ascii chars are expanded in routing algorithm so: `/tést` become: `/tést` and `t%C3%A9st`
- fix #3070
- fix #3071
cc/ @ogrammer


ogrammer:
Thank you, regarding your fix for static assets compared to mine, I think this one fits better. Mine was to decode the URI before checking the filesystem. And I see that you've made a new release! Thank you once more, it's really appreciated.

ogrammer:
Bug found: #3084

Also, is it intentional that consumer.accept is called 2 times for non-ASCII routes?

jknack:
@ogrammer we probably don't need the double call. I just keep for internal/server use only. So for example you can do things like: `ctx.matches("/tést")` with no need to encode the path pattern

PR #3134, opened 2023-09-12
Improve asset API

agentgt:
This will make it impossible for zero asset sources.

In theory this will break semver API but I know this is the right way and we don't follow semver :P 

Ideally this would be on the AssetHandler as well.

For the odd place where someone actually passes an array you can add an additional method:

```java
default @NonNull Route assets(@NonNull String pattern, @NonNull AssetSource[] sources)
```

Let me know and I can add that as well.


jknack:
> In theory this will break semver API but I know this is the right way and we don't follow semver :P

hahahah

I saw it before... didn't write it down and then forgot.

jknack (approved):

PR #3171, opened 2023-10-11
Cleaned up "Getting Started" docs

MrRawbin:
Thanks for a very cool project!

I am currently evaluating Jooby for my workplace, and when reading through the docs (which I believe are generally good) I noticed some grammatical errors. I think it would be a shame if these errors were to cast a shadow over such a cool framework, so I cleaned up the "Getting Started" part of the docs as a first effort. If you like these changes, I can go through the rest of the docs later if you'd like.


jknack (approved):
Thank you. I'll wait for more.

PR #3215, opened 2023-10-30
Add media types for DOCX, XLSX, PPTX

extbe:
- DOCX https://www.iana.org/assignments/media-types/application/vnd.openxmlformats-officedocument.wordprocessingml.document
- XLSX https://www.iana.org/assignments/media-types/application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
- PPTX https://www.iana.org/assignments/media-types/application/vnd.openxmlformats-officedocument.presentationml.presentation


jknack:
Hi,

branch 2.x isn't open to new addition.

PR #3233, opened 2023-11-10
Test JDK 21 instead of 20

SentryMan:
20 is EOL in most places


jknack:
yea, I tried it before but kotlin is fails to build on 21

PR #3302, opened 2023-12-22
Add @context support for JStachio module

agentgt:
@jknack 

JStachio added support for loose key->value stuff like request attributes. Originally this was a pain in the butt if you had say CSRF token you would have to add it manually to all models.

More info on that can be found here:

https://jstach.io/jstachio/io.jstach.jstachio/io/jstach/jstachio/context/package-summary.html

So now say something like CSRF token is bound to `Context.getAttributes` you can access it in JStachio like:

```
{{@context.csrf}}
```

This PR makes that possible.


jknack (commented):
Can you add a minimal example to the jstachio jooby doc?

agentgt:
@jknack Will do. I ran out of time right before the holidays and meant to leave it in draft mode for a little longer.

jknack:
@agentgt Ok, going to merge and release. We can update doc later.

Issue #3400, opened 2024-04-02
decoders not shared to the mounted routes


kliushnichenko:
Unlike jooby2 in 3.x decoders installed in the main app is not applied to the routes were mounted from another app

```java
class AppA extends Jooby {
    {
        post("/pets", ctx -> ctx.body(Pet.class));
    }
}

class MainApp extends Jooby {
    {
       install(new JacksonModule());
       mount(new AppA());
    }
  }
```

Now hit `post("/pets")` and you will catch `UnsupportedMediaType` error, cause Jackson decoder was overridden at the copy route stage.
Hope it wasn't changed intentionally and we can return the Jooby2 behaviour

jknack:
there is a slightly difference between `mount(Router)` and `install(Router)`. In your case `install` should work. Mount only import routes into the main app.

kliushnichenko:
yeah, indeed `install` works as expected, thank you

PR #3401, opened 2024-04-02
Fix decoders on mounted apps

kliushnichenko:
closes https://github.com/jooby-project/jooby/issues/3400


jooby/src/main/java/io/jooby/internal/RouterImpl.java
```diff
@@ -909,7 +909,7 @@ private void copy(Route src, Route it) {
     it.setTags(src.getTags());
     it.setDescription(src.getDescription());
     it.setDecoders(src.getDecoders());
-    it.setMvcMethod(it.getMvcMethod());
+    it.setMvcMethod(src.getMvcMethod());
```
kliushnichenko, line 912: @jknack this line still look like a type, should I close the PR or you are going to merge?

jknack: good catch! keep it, will merge later

PR #3411, opened 2024-04-18
Fix/dagger mvc openapi

kliushnichenko:
@jknack    
My previous PR https://github.com/jooby-project/jooby/pull/3398 about Avaje BeanScope controllers parsing in openapi broke the dagger controllers parsing. My apologies, here is the fix


edgarespinawt (changes requested):
cool, can you add a test like:

```java
{
   mvc(require(Bean.class));
}
```

don't remember if that works or not

kliushnichenko:
nope, `require` didn't work.  
added a test and fixed it

edgarespinawt (approved):

jknack (approved):

Issue #3412, opened 2024-04-18
[openapi] @QueryParam witn @NonNull annotation breaks openapi parser


kliushnichenko:
```java
import edu.umd.cs.findbugs.annotations.NonNull;

class Controller {

        @GET
        public void get(@QueryParam @NonNull String param1, 
                                 @QueryParam String param) { // it is important to have 2 or more params - then issue arise
            ...
        }
    }
```

The signature above leads to openapi generator to fail with the exception below

```text
Caused by: java.lang.NullPointerException: Cannot invoke "java.util.List.stream()" because "annotations" is null
	at io.jooby.internal.openapi.AnnotationParser.isNullable(AnnotationParser.java:449)
	at io.jooby.internal.openapi.AnnotationParser.routerArguments(AnnotationParser.java:386)
	at io.jooby.internal.openapi.AnnotationParser.routerMethod(AnnotationParser.java:259)
	at io.jooby.internal.openapi.AnnotationParser.parse(AnnotationParser.java:233)
	at io.jooby.internal.openapi.AnnotationParser.parse(AnnotationParser.java:208)
	at io.jooby.internal.openapi.RouteParser.routeHandler(RouteParser.java:334)
	at io.jooby.internal.openapi.RouteParser.parse(RouteParser.java:281)
	at io.jooby.internal.openapi.RouteParser.parse(RouteParser.java:79)
	at io.jooby.openapi.OpenAPIGenerator.generate(OpenAPIGenerator.java:177)
	at io.jooby.openapi.OpenAPIExtension.resolveParameter(OpenAPIExtension.java:58)
```

PR #3413, opened 2024-04-18
Fix/openapi nonnull issue

kliushnichenko:
Closes https://github.com/jooby-project/jooby/issues/3412   
For some reason `method.invisibleParameterAnnotations` contains `null` entries. It looks like ASM issue? I'm not good at it. But a dumb fix is here.


jknack:
Hi @kliushnichenko can you rebase?

kliushnichenko:
should be ok now

Issue #3415, opened 2024-04-24
How to expose Jooby module as Avaje injected bean


ludmiloff:
Hello,
I'm looking for a more complex Avaje usage, such as building ```@Singleton``` repositories, services etc. which are supposed to access a database or some other Jooby module.

Jooby itself brings its own kind-of service / dependency registration. The GuiceModule integrates with Jooby very well. Unfortunately, this is not the case with Avaje, which is completely not aware of Jooby  internals.

In particular, I've managed a workaround (see source below) by declaring the App class being a ```@Factory```. It works for a simple use case, despite it seems not optimal to me. 

```
@Factory
public class App extends Jooby {

  private BeanScope beanScope;

  public App(BeanScope beanScope) {
    super();
    this.beanScope = beanScope;

    install(new HikariModule());
    install(new JdbiModule());
    install(new NettyServer());

    mvc(this.beanScope.get(MyController.class));
  }

  @Bean
  Jdbi getJdbi() {
    return require(Jdbi.class);
  }

  @Bean
  Config appConfig() {
    return getConfig();
  }
}
```

```
public class Main {
    public static void main(final String[] args) {
        BeanScope beanScope = BeanScope.builder().build();
        runApp(args, () -> new App(beanScope));
    }
}
````

Where I stuck is QuartzModule with some jobs being declared as ```@Singleton```.
I'm really looking for some more idiomatic way, please advice.

**Comment by SentryMan** [comment]
I'd do it more like this:

```java
@Factory
public class App extends Jooby {

  public App(MyController controller) {
    install(new HikariModule());
    install(new JdbiModule());
    install(new NettyServer());

    mvc(controller);
  }

  @Bean
  Jdbi getJdbi() {
    return require(Jdbi.class);
  }

  @Bean
  Config appConfig() {
    return getConfig();
  }
}
```
but yeah I can work on adding a sort of Avaje Inject Module.

PR #3416, opened 2024-04-24
Add Avaje Inject Module

SentryMan:
First pass at an avaje inject module

helps alleviate #3415


modules/jooby-avaje-inject/src/main/java/io/jooby/avaje/inject/JoobyPropertyPlugin.java
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  @Override
+  public boolean contains(String property) {
+    return config.hasPath(property);
+  }
+
+  @Override
+  public boolean equalTo(String property, String value) {
+    return config.getString(property).equals(value);
```
edgarespinawt, line 29: should we check for existence first?

edgarespinawt (approved):
Good addition, thank you. Just one comment

ludmiloff (commented):
I grabbed the source and tried locally. It results in multiple Hikari re-initialisation and database connections and application crash. This is due to ```beanScope.provideDefault(key.getName(), key.getType(), e::getValue);``` which adds the bean as secondary. 
I suggest changing it to ```beanScope.bean(key.getType(), e.getValue());```.

With this change my testing code works, still needs ```@InjectModule(requires)``` everywhere.

ludmiloff (changes requested):
Please, change service registration as follows:
```java
if (key.getName() == null) {
  beanScope.provideDefault(key.getType(), e::getValue);
} else {
  beanScope.bean(key.getName(), key.getType(), e.getValue());
}
```

This:
1) prevents adding multiple secondary beans for same type, Avaje complains about that
2) allows named injection, e.g. ```@Named("db1") Jdbi``` 

ps: adding package-info.java with specific ```InjectModule(name = 'jooby-modules', requires = {...})``` is required for pass avaje-generator compillation

SentryMan:
@ludmiloff how's this?

ludmiloff (changes requested):
As @jknack mentioned before, there must be an existence check in JoobyPropertyPlugin, line 30
@SentryMan, may I suggest the change like this:
```java
return config.hasPath(property) && Objects.equals(config.getString(property), value);```

File: modules/jooby-avaje-inject/src/main/java/io/jooby/avaje/inject/AvajeInjectModule.java [hunk 2]
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * <p>Jooby integrates the {@link io.jooby.ServiceRegistry} into the Avaje DI framework.
+ *
+ * <p>Usage:
+ *
+ * <pre>{@code
+ * {
+ *
+ *
+ *   install(new AvajeInjectModule());
```
jknack, line 29: should be: `install(AvajeInjectModule.of())` or we need to make constructor public

modules/jooby-avaje-inject/src/main/java/io/jooby/avaje/inject/AvajeInjectModule.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  AvajeInjectModule(BeanScopeBuilder beanScope) {
+    this.beanScope = beanScope;
+  }
+
+  @Override
+  public void install(Jooby application) throws Exception {
+
+    application
```
jknack, line 59: This probably better to run on `starting` or `started`. So we don't depend of sequential order of services being added before this module.

SentryMan: would lateInit=true resolve?

ludmiloff (commented):
@SentryMan, your latest commit about named property injection works in the following scenario:

```java
@Inject @Named("myProp") String myProp
```

and the actual value is available during/after PostConstruct phase.

Constructor injection of named config item (Jooby 1 style, still good :) ) does not work, however.

Existence check in JoobyPropertyPlugin is fine.

ludmiloff (commented):
A bit of explanation about my latest comment:
constructor injection if named config item does not work because it needs to be declared as external dependency, which might be very cumbersome.

SentryMan:
you mean with the @InjectModule(requires) thing?

modules/jooby-avaje-inject/src/main/java/io/jooby/avaje/inject/AvajeInjectModule.java
ludmiloff: Tested, with lateinit=true the order does not matter anymore.

ludmiloff:
> you mean with the `@InjectModule(requires={String.class})` thing? For that, you only need to say `String.class` once and it'll take care of things. I'll need to make a note of that in the docs.

Yes, this way it works, thanks.

modules/jooby-avaje-inject/src/main/java/io/jooby/avaje/inject/AvajeInjectModule.java
jknack: right. lateinit will fix it

ludmiloff (commented):
I think this module is already done, at least for a sort of EAP. 
My notes about:
1. Needs documentation
2. Injecting property (incl. com.typesafe.config.Config items by name) works
3. Constructor injection works
4. All JoobyModule(s) currently in use are injected as Avaje Beans, named injection is also supported
5. All Avaje Beans are accessible via Jooby require
6. Installing AvajeInjectModule does not need a specific order in code along other Jooby modules
7. Packaging my test project into a single jar produces a working deployment without any problem.

Some limitation and caveats:
1. Using this module needs a package-info.java with proper InjectModule(require) clause. This is very Avaje specific.
2. Config item injection is currently limited to String types or string representation (toString). Might be improved with Integer and Boolean classes, though.
3. Using this module needs a special maven/gradle plugin and setup for Avaje code generation
4. Using this module adds a small overhead in startup time, on my testing machine 8 cores M1, java 17, small size project with 3 other jooby modules and 200 lines of config items, cold start in dev mode - around 20ms for wiring beans at runtime (as reported by Avaje), hope this is acceptable and much better than Guice

Finally, this is ready for approval and merging, IMO.
@jknack, @SentryMan, thank you for your support.

SentryMan:
> 3. Using this module needs a special maven/gradle plugin and setup for Avaje code generation

This should only be required if you are running on the module-path and also happen to have custom avaje autoconfig libraries. But yeah it is what it is

PR #3426, opened 2024-05-17
Fix avaje module

kliushnichenko:
@jknack 
AvajeInjectModule seems doesn't work as expected, at least for me.
With a very basic example (reproduced in test https://github.com/jooby-project/jooby/commit/c9c81fdc85470df43b086abb257d611a6fd0574a) Jooby inner services injection fails with class cast exception, for example `JsonMapper`:  
```text
Caused by: java.lang.ClassCastException: class io.jooby.internal.ServiceRegistryImpl$$Lambda$458/0x0000000800dbf350 cannot be cast to class com.fasterxml.jackson.databind.json.JsonMapper (io.jooby.internal.ServiceRegistryImpl$$Lambda$458/0x0000000800dbf350 and com.fasterxml.jackson.databind.json.JsonMapper are in unnamed module of loader 'app')
```

I assume this is due to the fact that services already wrapped into lambda in `singletone()` https://github.com/jooby-project/jooby/blob/03bb5875786f9414c9339902cf015e264d06e173/jooby/src/main/java/io/jooby/internal/ServiceRegistryImpl.java#L61

Hence,  the default binding need to be changed to
`beanScope.provideDefault(key.getType(), e.getValue()::get);`

Also, `Config` binding was removed as it is already in service registry

### DIFF
--- a/modules/jooby-avaje-inject/src/main/java/io/jooby/avaje/inject/AvajeInjectModule.java
+++ b/modules/jooby-avaje-inject/src/main/java/io/jooby/avaje/inject/AvajeInjectModule.java
@@ -80,7 +80,6 @@ public class AvajeInjectModule implements Extension {
     // configuration properties
     final var config = environment.getConfig();
     beanScope.propertyPlugin(new JoobyPropertyPlugin(config));
-    beanScope.bean(Config.class, config);
 
     for (var entry : config.entrySet()) {
       String name = entry.getKey();
--- a/modules/jooby-avaje-inject/src/main/java/io/jooby/avaje/inject/AvajeInjectModule.java
+++ b/modules/jooby-avaje-inject/src/main/java/io/jooby/avaje/inject/AvajeInjectModule.java
@@ -8,8 +8,6 @@ package io.jooby.avaje.inject;
 import java.util.List;
 import java.util.stream.Collectors;
 
-import com.typesafe.config.Config;
-
 import io.avaje.inject.BeanScope;
 import io.avaje.inject.BeanScopeBuilder;
 import io.jooby.Extension;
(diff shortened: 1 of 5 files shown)


modules/jooby-avaje-inject/src/main/java/io/jooby/avaje/inject/AvajeInjectModule.java
```diff
@@ -68,7 +66,7 @@ public void install(Jooby application) throws Exception {
             e -> {
               final var key = e.getKey();
               if (key.getName() == null) {
-                beanScope.provideDefault(key.getType(), e::getValue);
+                beanScope.provideDefault(key.getType(), e.getValue()::get);
```
jknack, line 69: Thanks

modules/jooby-avaje-inject/src/test/java/io/jooby/avaje/inject/app/Controller.java
```diff
@@ -0,0 +1,36 @@
...
+import io.avaje.inject.InjectModule;
+import io.jooby.annotation.GET;
+import io.jooby.annotation.Path;
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+import jakarta.inject.Singleton;
+
+@Singleton
+@Path("")
+@InjectModule(requires = {JsonMapper.class, Config.class, String.class})
```
jknack, line 14: never use Avaje Inject. Why do we need this line?

kliushnichenko:
```text
In Avaje Inject, the dependency graph is typically validated when the application compiles. As beans provided by Jooby Modules are registered at runtime, you must add @InjectModules(requires={String.class, <other Jooby Module Provided Bean ClassNames>…​}) to inform the avaje processor that these beans are provided at runtime.
```
from https://jooby.io/#dependency-injection-avaje-inject-mvc-routes :)

PR #3427, opened 2024-05-18
Fix #3425 jooby-pac4j module: set config for ProfileManager in Pac4jC…

### LINKED ISSUE #3425: jooby-pac4j oidc: never renew profile?
Hello,

I'm having an issue with renewing the profile after  has expired (for OIDC flow).

The problem arises because the profile is cleared every time in the ProfileManager, as the class has not set the config. Consequently, the profile is never renewed.

Take a look at the code:
In [Pac4JModule](https://github.com/jooby-project/jooby/blob/3.x/modules/jooby-pac4j/src/main/java/io/jooby/pac4j/Pac4jModule.java#L584) the current user is set as follows:
```java
application.setCurrentUser(new Pac4jCurrentUser());
```

The current user class has the following method:
```java
  public Object apply(Context ctx) {
    Pac4jContext pac4jContext = Pac4jContext.create(ctx);
    ProfileManager pm = new ProfileManager(pac4jContext, pac4jContext.getSessionStore());
    return pm.getProfile().orElse(null);
  }
```
Here, we have an initialized [ProfileManager](https://github.com/pac4j/pac4j/blob/5.7.x/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java) without a config. The config should probably be set using the public method of ProfileManager setConfig(Config config).

The getProfile method calls the ``removeOrRenewExpiredProfiles`` method in ``ProfileManager``, which includes the following condition:
```java
    protected void removeOrRenewExpiredProfiles(final LinkedHashMap<String, UserProfile> profiles, final boolean readFromSession) {
        var profilesUpdated = false;
        for (final var entry : profiles.entrySet()) {
            final var key = entry.getKey();
            final var profile = entry.getValue();
            if (profile.isExpired()) {
                LOGGER.debug("Expired profile: {}", profile);
                profilesUpdated = true;
                profiles.remove(key);
                if (config != null && profile.getClientName() != null) {
                    final var client = config.getClients().findClient(profile.getClientName());
                    if (client.isPresent()) {
                        try {
                            final var newProfile = client.get().renewUserProfile(profile, context, sessionStore);
                            if (newProfile.isPresent()) {
                                LOGGER.debug("Renewed by profile: {}", newProfile);
                                profiles.put(key, newProfile.get());
                            }
                        } catch (final RuntimeException e) {
                            logger.error("Unable to renew the user profile for key: {}", key, e);
                        }
                    }
                }
            }
        }
        if (profilesUpdated) {
            saveAll(profiles, readFromSession);
        }
    }
```
In the method you see the condition
```java
if (config != null && profile.getClientName() != null) {
```
I suspect that this condition will always be null when ProfileManager is called by Pac4jCurrentUser, resulting in the profile always being cleared.

Could you please advise me on how to bypass this problem? Or perhaps it should be fixed?

Thank you.

tkarlinski:
Fix renewing the Profile in the jooby-pac4j module by setting the configuration for the ProfileManager in Pac4jCurrentUser.

### DIFF
--- a/modules/jooby-pac4j/src/main/java/io/jooby/pac4j/Pac4jModule.java
+++ b/modules/jooby-pac4j/src/main/java/io/jooby/pac4j/Pac4jModule.java
@@ -580,8 +580,11 @@ public class Pac4jModule implements Extension {
             options.setDefaultUrl(defaultUrl);
           });
     }
+
+    application.getServices().put(Config.class, pac4j);
+
     /** Set current user provider */
-    application.setCurrentUser(new Pac4jCurrentUser(pac4j));
+    application.setCurrentUser(new Pac4jCurrentUser());
     // cleanup
     clientMap.clear();
   }
--- a/modules/jooby-pac4j/src/main/java/io/jooby/pac4j/Pac4jModule.java
+++ b/modules/jooby-pac4j/src/main/java/io/jooby/pac4j/Pac4jModule.java
@@ -581,7 +581,7 @@ public class Pac4jModule implements Extension {
           });
     }
     /** Set current user provider */
-    application.setCurrentUser(new Pac4jCurrentUser());
+    application.setCurrentUser(new Pac4jCurrentUser(pac4j));
     // cleanup
     clientMap.clear();
   }
--- a/modules/jooby-pac4j/src/main/java/io/jooby/internal/pac4j/Pac4jCurrentUser.java
+++ b/modules/jooby-pac4j/src/main/java/io/jooby/internal/pac4j/Pac4jCurrentUser.java
@@ -8,15 +8,24 @@ package io.jooby.internal.pac4j;
 import java.util.function.Function;
 
 import org.pac4j.core.profile.ProfileManager;
+import org.pac4j.core.config.Config;
 
 import io.jooby.Context;
 import io.jooby.pac4j.Pac4jContext;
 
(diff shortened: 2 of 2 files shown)


modules/jooby-pac4j/src/main/java/io/jooby/pac4j/Pac4jModule.java
```diff
@@ -581,7 +581,7 @@ public void install(@NonNull Jooby application) throws Exception {
           });
     }
     /** Set current user provider */
-    application.setCurrentUser(new Pac4jCurrentUser());
+    application.setCurrentUser(new Pac4jCurrentUser(pac4j));
```
jknack, line 584: Let's make the pac4j configuration accesible by adding to registry:

```
application.getServices().put(Config.class, pac4j);
```

jknack (changes requested):
Looks good, let's make one more change.

modules/jooby-pac4j/src/main/java/io/jooby/pac4j/Pac4jModule.java
tkarlinski: OK. I added the configuration to the registry at the end of the ``install`` method because above, the configuration can be changed

jknack (approved):

PR #3502, opened 2024-08-14
415 unsupported media type when using mount to add routes

### LINKED ISSUE #3400: decoders not shared to the mounted routes

### LINKED ISSUE #3500: 415 unsupported media type when using mount to add routes
This is a weird one that started happening since 3.0.8 onwards.

If I add a POST route that is declared in another router via a `mount` then I get a 415 error when POSTing a valid JSON payload.

If I add the same route directly to the server, I get the expected valid response.

Best demonstrated via the following MWE:

```java
import io.jooby.Jooby;
import io.jooby.StatusCode;
import io.jooby.gson.GsonModule;

public class WidgetService extends Jooby {

	public WidgetService() {
		install(new GsonModule());

		post("/api/widgets1", ctx -> {
			Widget widget = ctx.body().to(Widget.class);
			System.out.println("Created " + widget);
			return ctx.send(StatusCode.CREATED);
		});

		mount(new WidgetRouter());

	}

		new WidgetService().start();
	}

}

class WidgetRouter extends Jooby {

	public WidgetRouter() {

		post("/api/widgets2", ctx -> {
			Widget widget = ctx.body().to(Widget.class);
			System.out.println("Created " + widget);
			return ctx.send(StatusCode.CREATED);
		});

	}
}

```
`Widget` is just a POJO with a couple of string fields.

A POST containing valid JSON to `/api/widgets1` works and returns the expected 201 response.

The same POST to `/api/widgets2`  throws a 415 unsupported media type with the following stack trace:

```
ERROR WidgetService - POST /api/widgets2 415 Unsupported Media Type
 io.jooby.exception.UnsupportedMediaType: application/json
	at io.jooby.MessageDecoder.lambda$static$0(MessageDecoder.java:24)
	at io.jooby.DefaultContext.decode(DefaultContext.java:420)
	at io.jooby.internal.netty.NettyBody.to(NettyBody.java:103)
	at io.jooby.Body.to(Body.java:100)
	at WidgetRouter.lambda$new$baa66101$1(WidgetService.java:32)
	at io.jooby.internal.handler.DefaultHandler.lambda$apply$e67b40fd$1(DefaultHandler.java:21)
	at io.jooby.Route$Filter$1.apply(Route.java:94)
	at io.jooby.internal.handler.WorkerHandler.lambda$apply$0(WorkerHandler.java:22)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

I'm using the following dependencies in my `build.gradle`:

```
dependencies {
	def joobyVer = '3.2.7';
	implementation group: 'io.jooby', name: 'jooby-netty', version: joobyVer;
	implementation group: 'io.jooby', name: 'jooby-gson', version: joobyVer;
	implementation group: 'ch.qos.logback', name: 'logback-classic', version: '1.5.6';
}
```

I've been messing around with this for a bit trying to work out why my service started breaking after I upgraded from 3.0.7 to 3.2.7, and as far as I can tell, it is the use of `mount` to add the route that leads to the 415 error.

I've tried Jackson instead of Gson, and get the same result.

As mentioned, versions prior to 3.0.8 don't seem to have this problem.

jknack:
- fix #3500
- fix #3400

### DIFF
--- a/jooby/src/main/java/io/jooby/internal/RouterImpl.java
+++ b/jooby/src/main/java/io/jooby/internal/RouterImpl.java
@@ -911,7 +911,8 @@ public class RouterImpl implements Router {
     it.setExecutorKey(src.getExecutorKey());
     it.setTags(src.getTags());
     it.setDescription(src.getDescription());
-    it.setDecoders(src.getDecoders());
+    // DO NOT COPY: See https://github.com/jooby-project/jooby/issues/3500
+    // it.setDecoders(src.getDecoders());
     it.setMvcMethod(src.getMvcMethod());
     it.setNonBlocking(src.isNonBlocking());
     it.setSummary(src.getSummary());
@@ -983,7 +984,7 @@ public class RouterImpl implements Router {
       Jooby app = (Jooby) router;
       override(src, app.getRouter(), consumer);
     } else if (router instanceof RouterImpl that) {
-      consumer.accept((RouterImpl) src, that);
+      consumer.accept(src, that);
     }
   }
 
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -35,6 +35,11 @@
     </dependency>
     <dependency>
       <groupId>io.jooby</groupId>
+      <artifactId>jooby-gson</artifactId>
+      <version>${jooby.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.jooby</groupId>
       <artifactId>jooby-avaje-jsonb</artifactId>
       <version>${jooby.version}</version>
     </dependency>
@@ -250,7 +255,6 @@
       <version>1.37</version>
       <scope>test</scope>
     </dependency>
-
   </dependencies>
 
   <build>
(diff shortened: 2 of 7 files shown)


kliushnichenko:
looks good, if build failures are not related to the changes 😄

PR #3508, opened 2024-08-24
Feat/jooby hibernate validator

kliushnichenko:
Still need some docs, tests, cleanup, but the basic logic seems done


agentgt:
@kliushnichenko @jknack 

I'm not sure I like the name of jooby-validation if it is going to have a hard dependency on jakarta.validation. Is the plan to make the validation pluggable in that module and it is just jakarta for now?

Otherwise I think it should be jooby-jakarta-validation.


Also the jooby-apt module should not have any dependency on basically anything ideally. I haven't code dived to see why you made it that way. (you don't need the actual classes you depend on with generated code)

modules/jooby-apt/pom.xml
```diff
@@ -19,6 +19,12 @@
 
   <dependencies>
 
+    <dependency>
```
jknack, line 22: As @agentgt suggested we don't need the dependency.

modules/jooby-validation/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+    <artifactId>jooby-validation</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.jooby</groupId>
+            <artifactId>jooby</artifactId>
+            <version>${jooby.version}</version>
+        </dependency>
+
+        <dependency>
```
jknack, line 22: @kliushnichenko @agentgt we can probably add the dependency on `jooby` (core). We have the inject already. This also means remove/delete this new module and move the ValidationHelper to a jooby core package.

kliushnichenko: The idea was to keep the core free of any optional dependencies, since not every application needs/uses validation. So, it might be better to avoid adding dependencies that are rarely used.

`jakarta-inject` is tiny comparing to `validation`, and DI I believe is much more popular.
Keep in mind except `ValidationHelper` there will be ExceptionHandler added as well to the core.

Do you still want to have it in the core? You are the boss, just want to make everything clear

jknack: > jakarta-inject is tiny comparing to validation

good point. Let's keep the new module, but let's rename to `jooby-jakarta-validation`

modules/jooby-apt/pom.xml
kliushnichenko: About this dependency. You are right it is not necessary to successfully build/compile the module. But, the generated code might rely on it, and someone should bring it into the runtime, cause `jooby-apt` doesn't. It is some sort of a hidden dependency. I decided to highlight this moment at least this way.

We can change the scope to `test` or remove it completely and add some notes in the internal documentation, dunno which option is better.

SentryMan:
The way I've done it with other libs I've worked was to have an extensible [Validator](https://github.com/avaje/avaje-http/blob/master/http-api/src/main/java/io/avaje/http/api/Validator.java) interface which clients can extend to provide custom validation. Example [hibernate impl](https://github.com/avaje/avaje-http/blob/master/http-hibernate-validator/src/main/java/io/avaje/http/hibernate/validator/BeanValidator.java) and [avaje validation impl ](https://github.com/avaje/avaje-validator/blob/main/validator-http-plugin/src/main/java/io/avaje/validation/http/BeanValidator.java). Perhaps you guys could try something similar if you want to eliminate the dependencies

modules/jooby-apt/pom.xml
jknack: Let's remove it to make jooby-apt works as it is today. We can check from validator if at least `jooby-hibernate-validator` is on classpath and produces an error when not

jknack:
@SentryMan thank you. I feel here isn't necessary to add another abstraction over the jakarta one. Leaving jakarta-validator outside of core is enough.

kliushnichenko:
Updated. Pls verify the changes and I'll proceed with the documentation

modules/jooby-apt/pom.xml
kliushnichenko: Actually, we still need it for tests, so, just changed the scope

kliushnichenko:
@SentryMan could you pls shed some light on `avaje-validator`. I thought it is fully compliant with `jakarta-validation-api`. But I can't find how I can get the instance of `jakarta.validation.Validator` baked by `avaje-validator`. All the entry points are available only at your custom abstraction `io.avaje.validation.Validator`.

Is it a trade-off of a reflection-free solution or maybe you can provide us jakarta-compliant entry point in the future release?

SentryMan:
> Is it a trade-off of a reflection-free solution

Jakarta validation is mainly reflection based yeah so we couldn't fully implement. Hence why we took a page from spring in our http solution by defining another abstraction so we could support both jakarta and avaje validation

modules/jooby-hbv/src/main/java/io/jooby/hbv/HbvModule.java
```diff
@@ -0,0 +1,96 @@
...
+import jakarta.validation.Validator;
+import jakarta.validation.ValidatorFactory;
+import org.hibernate.validator.HibernateValidator;
+import org.hibernate.validator.HibernateValidatorConfiguration;
+
+import java.util.function.Consumer;
+
+import static jakarta.validation.Validation.byProvider;
+
+public class HbvModule implements Extension {
```
jknack, line 24: I do prefer: `HibernateValidatorModule` same for maven coordinates: `jooby-hibernate-validator`

modules/jooby-hbv/src/main/java/module-info.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Jooby https://jooby.io
+ * Apache License Version 2.0 https://jooby.io/LICENSE.txt
+ * Copyright 2014 Edgar Espina
+ */
+/**
+ * Hbv module.
+ */
+module io.jooby.hbv {
```
jknack, line 9: same here: `io.jooby.hibernate.validator`

modules/jooby-jakarta-validation/src/main/java/io/jooby/validation/ConstraintViolationHandler.java
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Catches and transform {@link ConstraintViolationException} into {@link ValidationResult}
+ *
+ * Payload example:
+ * <code>
+ * {
+ *     "title": "Validation failed",
+ *     "status": 422,
+ *     "errors": {
+ *         "objectErrors": [
```
jknack, line 26: In gral, I do prefer a single object type where "field" will be nullable/missing. Like:

```json
"errors": [{
   "field": "name" or missing/null,
   "messages": []
   "type": "field" or "global" or "object"
}]
```

modules/jooby-jakarta-validation/src/test/java/io/jooby/validation/app/PasswordsShouldMatchValidator.java
```diff
@@ -0,0 +1,15 @@
+package io.jooby.validation.app;
+
+import jakarta.validation.ConstraintValidator;
+import jakarta.validation.ConstraintValidatorContext;
+
+public class PasswordsShouldMatchValidator implements ConstraintValidator<PasswordsShouldMatch, NewAccountRequest> {
+
+    @Override
+    public boolean isValid(NewAccountRequest request, ConstraintValidatorContext constraintContext) {
```
jknack, line 9: Is there a way to access to jooby registry from `ConstraintValidatorContext`?????

jknack (commented):
looking good

modules/jooby-jakarta-validation/src/main/java/io/jooby/validation/ConstraintViolationHandler.java
kliushnichenko: I tried to make the structure more UI-friendly. UI usually needs field/global errors in separate "baskets". On the other hand, the client is not necessarily UI, so, fair enough.

Let's clarify the types:
- field - it is clear
- global - I suppose it is for cross-field error on the top-level bean
- object - is it for nested beans? in this case, we might need to introduce one more property, cause right now the path for the nested bean errors looks like a full path to the field e.g. `person.firstName`. And in case of the list - `person[0].firstName`

jknack: It's the first time I see the objectErrors (didn't know about them until now). So I tried to create a fixed structure for clients (UI or not). Clients will mostly want to display the error message and probably the field name. Type is for those clients that want to know more about the error so based on your example type should be `field` (for fieldErrors) or `object` (for objectErrors). Does it make sense now?

Additionally we could render the `path` too which might or might not match `field` attribute

modules/jooby-jakarta-validation/src/test/java/io/jooby/validation/app/PasswordsShouldMatchValidator.java
jknack: > Additionally, if the project utilizes some DI it is not a problem to inject any service/repo/etc. into custom validator

An example of that on docs will be nice.

kliushnichenko:
Getting back to `avaje-validator` support...

As SentryMan clarified, there is no way to smoothly support both (hbv/avaje) relying solely on `jakarta.validation.Validator`
So, my initial plan didn't quite work out 😄 

Options that are running through my mind in this regard:

1. Write a custom implementation of the `jakarta.validation.Validator` that will trigger `io.avaje.validation.Validator` under the hood
**Concerns:**    
way too odd. Everything except one method `validate()` will be stubbed as `Unsupported`.   
Also, it will hide the rest of `avaje-validator` options, not ok.

2. Each module can implement its own helper for validation. In `jooby-apt` pick the one that is available in classpath.
**Concerns:**   
Complicates `jooby-apt`, probably duplicates some logic in two places, error-prone, complicates further support.

3. Introduce one more abstraction, as SentryMan pointed out, to cover `jooby-apt` logic over a single interface.
**Concerns:** 
plus one more abstraction, a bit more cumbersome, but, seems the most acceptable compromise 

@jknack what's your take on it? maybe some other options?

jknack:
Seems 3 it is OK. jooby-apt should call io.jooby.validation.BeanValidator and extra logic goes here and probably on the jooby-avaje-validator module.

So jooby-apt will do what is doing here, wrap the call around BeanValidator

kliushnichenko:
Updated.

Please note the following:

- since `avaje-validator` has its own set of custom entities (`Validator`, `ConstraintViolation`, `ConstraintViolationException`), the `BeanValidator` helper no longer relies on either jakarta or avaje entities, making the module even simpler.
- for the same reason, error handlers should be implemented separately and placed in their corresponding modules.
- `jooby-jakarta-validation` has been renamed back to `jooby-validation` since it no longer has any dependency on jakarta.
- a new `MvcValidator` interface has been introduced to isolate `BeanValidator` logic from specific implementations. It includes a single `validate()` method that performs validation and throws an exception. The type of exception thrown depends on the implementation (Jakarta or Avaje).

Naming is getting tough with a lot of `Validator` abstractions, any suggestions are welcomed.  
@jknack pls confirm we a good with this solution and can move on

modules/jooby-apt/src/main/java/io/jooby/apt/JoobyProcessor.java
```diff
@@ -371,4 +372,21 @@ public static RuntimeException propagate(final Throwable x) {
...
+      TypeElement validatorElement = processingEnv.getElementUtils()
+              .getTypeElement("io.jooby.validation.BeanValidator");
+
+      if (validatorElement == null) {
+        processingEnv.getMessager().printMessage(
+                Diagnostic.Kind.ERROR,
+                "Unable to load 'BeanValidator' class. " +
+                "Bean validation usage (@Valid) was detected, but the appropriate dependency is missing. " +
+                "Please ensure that you have added the corresponding validation dependency " +
+                "(e.g., jooby-hbv).");
```
jknack, line 388: typo: `jooby-hibernate-validator`

modules/jooby-validation/src/main/java/io/jooby/validation/MvcValidator.java
```diff
@@ -0,0 +1,6 @@
+package io.jooby.validation;
+
+public interface MvcValidator<E extends Throwable> {
```
jknack, line 3: probably make E RuntimeException

modules/jooby-validation/src/main/java/io/jooby/validation/MvcValidator.java
```diff
@@ -0,0 +1,6 @@
...
+
+    void validate(Object bean) throws E;
```
jknack, line 5: well not sure if we need a generic for exception, RuntimeExceptoion or Exception will be OK

jknack (approved):
few minor comments

kliushnichenko:
minor improvements, javadoc, asciidoc has been added. Ready to final review/merge.  
avaje-validator-module I'll send in a separate PR

jknack (approved):
@kliushnichenko love what you did! Let me know if we can merge.

kliushnichenko:
Yes, we can. Just a minor thing, master and my branch are under `3.2.10-SNAPSHOT`, but I see `3.3.0` has already been released, is it expected, or we need to update both to `3.3.1-SNAPSHOT`?

jknack:
@kliushnichenko done

kliushnichenko:
@jknack updated, feel free to merge

PR #3517, opened 2024-09-05
Fixes case insensitive headerMap

froque:
Makes map returned by headerMap consistent with header() method.


froque:
Replaced by #3518

PR #3518, opened 2024-09-06
Fixes case insensitive headerMap and header toMultimap

froque:
Makes all ways to work with headers consistent and comply with the HTTP Headers RFC.

Continues from #2358 
Replaces #3517


jknack (approved):
Good fix!! Please run `mvn clean package` so code gets formatted properly.

PR #3520, opened 2024-09-07
validation: cleanup

jknack:
- make constrainctvalidation cast safe
- use config api for properties
@SentryMan can you look at locales?


SentryMan (approved):
LGTM, the validator should use the default system locale if none are provided

kliushnichenko:
@jknack Don't you want to add some `.editorconfig` to avoid manual code re-formatting? And maybe some contribution guide

jknack:
we should add it yea. Today it uses spotless + git pre-commit hook that is installed on `mvn clean package`

SentryMan:
I'm in the minority, but I use Eclipse, so `.editorconfig` doesn't do anything for me.

PR #3522, opened 2024-09-07
validation: handler/filter for lambda routes + remove MvcValidator

jknack:
- Remove `MvcValidator`. Merge content into `BeanValidator`
- Rename BeanValidator.validate to BeanValidator.apply (avoid method name conflict)
- Creates BeanValidator.validate() route handler/filter which intercept calls to HTTP request object and run validator
- TODO: if we agreed on merging of MvcValidator we can remove/get rid of `jooby-validation` module

There two ways of using validator on lambda routes:

```
{
    get("/do", validate(ctx -> ctx.query(Bean.class)));
}
```

Or with filter:

```
{
    use(validate());
    get("/do", ctx -> ctx.query(Bean.class));
}
```


jknack:
@SentryMan @kliushnichenko one more

kliushnichenko:
the developer experience is much better with a filter, but, the price is reflection/proxy usage, are we fine with it? 
I remember you were discussing GraalVM support, is it still on the roadmap?

jknack:
> GraalVM

Nothing more was done for long time.

Let me remove reflection. I'm just lazy sometimes.

jknack:
@kliushnichenko without reflection

kliushnichenko:
It's a bit late, but 
I realized that if the project shares both approaches (mvc/lambda)
and uses filter `use(validate())`
that means that MVC route beans will be validated twice...

Sharing mvc/lambda is probably not a common case, but still realistic/possible.
Maybe we should remove the filter, and keep only the handler, bc behavior is quite odd.
Or at least add a note in the documentation

jknack:
Feel it is an odd case (like you said), but we can probably do something to fix it. 

can you work and mentioned on documentation?

PR #3534, opened 2024-09-17
cleanup validation tests

kliushnichenko:
just a bit of cleaning

### DIFF
--- a/docs/asciidoc/context.adoc
+++ b/docs/asciidoc/context.adoc
@@ -576,7 +576,7 @@ a javadoc:Session[] but the lifecycle is shorter: *data is kept for only one req
   });
 
   post("/save", ctx -> {
-    ctx.flash("success", "Item created");          <1>
+    ctx.flash().put("success", "Item created");    <1>
     return ctx.sendRedirect("/");                  <2>
   });
 ----
@@ -589,7 +589,7 @@ a javadoc:Session[] but the lifecycle is shorter: *data is kept for only one req
   }
   
   post("/save") { ctx ->
-    ctx.flash("success", "Item created")           <1>
+    ctx.flash().put("success", "Item created")     <1>
     ctx.sendRedirect("/")                          <2>
   }
 ----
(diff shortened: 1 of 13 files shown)


docs/asciidoc/context.adoc
```diff
@@ -576,7 +576,7 @@ a javadoc:Session[] but the lifecycle is shorter: *data is kept for only one req
   });
 
   post("/save", ctx -> {
-    ctx.flash("success", "Item created");          <1>
```
kliushnichenko, line 579: just aligning the doc with what we have at the moment. There is no `ctx.flash(str,str)` method

modules/jooby-apt/src/test/java/tests/validation/BeanValidationsController.java
```diff
@@ -22,12 +22,6 @@ public Bean validateFormBean(@Valid @FormParam Bean bean) {
     return bean;
   }
 
-  // todo: revive when flash `toNullable` will be fixed
```
kliushnichenko, line 25: What I meant here is that it is possible to define the param as object `@FlashParam Bean bean`   
and apt will generate the code that is not executable, since `toNullable` is not there for flash/session/...
image

The use-case a bit synthetic, but still, I can add some type checking if you think it makes sense

jknack: Let's add the type check

jknack (changes requested):

kliushnichenko:
Added type verification.

Flash, Session, Cookie, and Header params are now restricted to `BUILT_IN` types.
Let me know if I missed smthg.

Error message will look like this:
```text
Illegal argument type at 'tests.verifyarg.ControllerFlash.param()'. 
Parameter 'param' annotated as [io.jooby.annotation.FlashParam] cannot be of type 'java.lang.Object'. 
Supported types are: [java.lang.Byte, java.lang.Short, java.lang.Character, java.lang.String, short, int, java.net.URI, java.lang.Float, java.lang.Boolean, float, char, java.time.LocalDateTime, byte, java.lang.Enum, java.math.BigInteger, java.util.UUID, java.util.Date, java.lang.Long, java.math.BigDecimal, io.jooby.StatusCode, java.lang.Double, boolean, java.net.URL, java.time.ZoneId, java.util.TimeZone, java.lang.Integer, java.time.LocalDate, java.time.Instant, java.time.Duration, java.nio.charset.Charset, java.time.Period, double, long]

```

jknack (approved):

PR #3552, opened 2024-10-02
Support Problem Details for HTTP APIs RFC7807

kliushnichenko:
This PR introduces:
1) `HttpProblem` class that represents the model of `RFC7807` problem details.
It extends `RuntimeException` so it can be naturally thrown.
It also provides handy static methods to instantiate a problem and a builder to make a fully packed instance.

2) `ProblemDetailsErrorHandler` a global error handler aimed to make anything thrown from the app compliant with `RFC7807`.
It also contains special handling for some built-in exceptions to make the response more informative.
It always logs server errors(5xx) and can optionally log client errors (4xx) unless they are fully muted

I'm going to align our `ValidationResult` with RFC7807 in a separate PR. That's why both classes were placed in the core package nevertheless it may seem odd at the moment.

If there are no critical comments I'll proceed to the documentation part.

@jknack pls take a look when you have a spare time


jooby/src/main/java/io/jooby/problem/ProblemDetailsErrorHandler.java
```diff
@@ -0,0 +1,218 @@
...
+import org.slf4j.Logger;
+
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import static io.jooby.MediaType.*;
+import static io.jooby.StatusCode.*;
+
+public class ProblemDetailsErrorHandler extends DefaultErrorHandler {
```
jknack, line 20: Let's rename to `ProblemDetailsHandler`

jooby/src/main/java/io/jooby/problem/ProblemDetailsErrorHandler.java
```diff
@@ -0,0 +1,218 @@
...
+
+  public ProblemDetailsErrorHandler log4xxErrors() {
+    this.log4xxErrors = true;
+    return this;
+  }
+
+  @Override
+  public void apply(@NonNull Context ctx, @NonNull Throwable cause, @NonNull StatusCode code) {
+    Logger log = ctx.getRouter().getLog();
+    if (cause instanceof NotAcceptableException ex) {
```
jknack, line 32: let's create an interface so exceptions can implement/convert to Problem

jooby/src/main/java/io/jooby/problem/ProblemDetailsErrorHandler.java
```diff
@@ -0,0 +1,218 @@
...
+    if (json.equals(type)) {
+      ctx.setResponseType(PROBLEM_JSON);
+    } else if (xml.equals(type)) {
+      ctx.setResponseType(PROBLEM_XML);
+    }
+  }
+
+  private HttpProblem evaluateTheProblem(Context ctx, Throwable cause, StatusCode statusCode) {
+    HttpProblem problem;
+    if (cause instanceof HttpProblem httpProblem) {
```
jknack, line 81: let's reuse here

jooby/src/main/java/io/jooby/problem/ProblemDetailsErrorHandler.java
kliushnichenko: If you need to create a custom problem you can do it by extending the `HttpProblem`. Here is an example https://github.com/kliushnichenko/jooby/blob/feat/problem-details/tests/src/test/java/io/jooby/problem/data/OutOfStockProblem.java

It is quite convenient to reuse the builder and it gives us enough control to guarantee that the problem instance is legit.
I'm afraid with an interface it won't be so convenient, and it for sure lacks control (for example the title is required and the status code should be valid). Also, static helper methods need to be moved somewhere...

It looks like there are more cons than pros. Or maybe I didn't get your point correctly

jooby/src/main/java/io/jooby/problem/ProblemDetailsErrorHandler.java
kliushnichenko: following the answer above, it will catch problems inherited from `HttpProblem` appropriately

jooby/src/main/java/io/jooby/problem/ProblemDetailsErrorHandler.java
jknack: I mean: 

```java

interface ToHttpProblem {
    HttpProblem toHttpProblem();
}
```

Probably a better name. So any exception can be converted to a HttpProblem (no need of custom message on handler) and when not we fallback to `HttpProblem.valueOf`

tests/src/test/java/io/jooby/problem/ProblemDetailsErrHandlerTest.java
```diff
@@ -0,0 +1,340 @@
...
+import java.util.Map;
+
+import static io.jooby.MediaType.PROBLEM_JSON;
+import static io.jooby.MediaType.html;
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.CoreMatchers.nullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
+
+@JoobyTest(value = App.class, port = 9099)
```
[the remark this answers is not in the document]
kliushnichenko: question. How much is it important to keep `BUILD_PORT: 0` for `CI` pipelines?  
I've switched to `@ServerTest`, but I'd like to remain with rest-assured (find it quite convenient for intg tests).
Unfortunately `rest-assured` requires port to be greater than 0. May I change it?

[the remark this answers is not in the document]
jknack: The real port is available here: https://github.com/jooby-project/jooby/blob/3.x/tests/src/test/java/io/jooby/junit/ServerTestRunner.java#L116 after we call `options.setPort`.

Let's add a field/getter to `ServerTestRunner` to get back the computed port.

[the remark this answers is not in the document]
kliushnichenko: quick solution for now, guess better support of rest-assured in `ServerTestRunner ` can be added later

jknack (approved):

jknack:
now I merged it, but sure if there was more work coming??

PR #3557, opened 2024-10-10
problem details in modules

kliushnichenko:
While was working on problem details support in validation modules the general concept a bit changed.  
Enabling the `ProblemDetailsHandler` moved to the `hocon` configuration.

1. It greatly simplifies the integration of problem details in modules. Modules can now detect if `ProblemDetailsHandler` is enabled at an early stage of the module initiation. With the previous approach, it would have been much harder and might have led to implementing awkward solutions.

2. An additional benefit of the new approach is that `ProblemDetailsHandler` will be automatically registered as the last handler in the error handlers chain, so dev no longer need to worry about that.

So, to enable `ProblemDetailsHandler` the next config now required:
```hocon
problem.details {
  enable: true
  muteCodes: [401, 106]
  muteTypes: ['com.example.MyException']
}
```

@jknack pls share your thoughts, if you are Ok with the new approach I'll continue with docs update and some minor improvements

### DIFF
--- a/docs/asciidoc/problem-details.adoc
+++ b/docs/asciidoc/problem-details.adoc
@@ -10,51 +10,43 @@ Thankfully, there’s a standard called https://www.rfc-editor.org/rfc/rfc7807[I
 By adopting `RFC 7807`, API designers don’t have to spend time creating a custom solution, and users benefit by recognizing a familiar format across different APIs.
 If it suits the API’s needs, using this standard benefits both designers and users alike.
 
-`Jooby` provides built-in support for `Problem Details`. There are two main entities to work with:
+`Jooby` provides built-in support for `Problem Details`.
 
-1. `HttpProblem` - representation of the `RFC 7807` model and the way to instantiate the problem.
+=== Set up ProblemDetails
 
-2. `ProblemDetailsHandler` - global error handler that catches all exceptions, transforms them into `Problem Details` compliant format and renders the response based on the `Accept` header value.
-It also sets the appropriate content-type in response (e.g. application/problem+json, application/problem+xml)
+To enable the `ProblemDetails`, simply add the following line to your configuration:
 
-=== Set up handler
-
-The bare minimal requirement is to set up an error handler:
-
-.Java
-[source,java,role="primary"]
+.application.conf
+[source, properties]
 ----
-import io.jooby.problem.ProblemDetailsHandler;
+problem.details.enabled = true
+----
 
-{
-  ...  
-  error(new ProblemDetailsHandler()          // <1>
-    .log4xxErrors()                          // <2>
-    .mute(StatusCode.UNAUTHORIZED)           // <3>
-  );
+This is the bare minimal configuration you need.
+It enables a global error handler that catches all exceptions, transforms them into Problem Details compliant format and renders the response based on the Accept header value. It also sets the appropriate content-type in response (e.g. application/problem+json, application/problem+xml)
+
+All supported settings include:
+
+.application.conf
+[source, properties]
+----
+problem.details {
+  enabled = true
+  log4xxErrors = true                                // <1>
+  muteCodes = [401, 403]                             // <2>
+  muteTypes = ["com.example.MyMutedException"]       // <3>
 }
 ----
 
-.Kotlin
-[source,kt,role="secondary"]
-----
-import io.jooby.problem.ProblemDetailsHandler
 
-{
-  ...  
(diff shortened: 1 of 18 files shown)


jknack:
I have fixed feelings. I like it but then is like we need to add the a new property in Jooby class isn't wrong but wonder if we could do it using an Extension + install call

```
install(new ProblemsDetailsModule());
```

Once installed take care of configure everything. The issue here is how to integrate with other error handler like validation

kliushnichenko:
Just for clarity, there is no new property in `Jooby` class, just a helper method that reads the config. We can move this helper to any other place, all it needs is access to the config.

Regarding module 
```java
install(new ProblemsDetailsModule());
```
If a new module only sets the same hocon config as it is now(just code-driven approach) it will work. But this module should always be registered before validation modules.

If the module should also register the handler `error(new ProblemDetailsHandler())`, that's the issue. 
Because the global handler should be the last in the chain. Even if we made this module `lateinit=true` it will fix one issue (global err handler now registered the last), but we introduce another issue - the config properties are set too late, and validation modules won't detect that ProblemDetails is enabled. 
So, the last option is to move the detection of ProblemDetails in validation modules to the stage when the first error is thrown - I believe this is the worst option. That's why I ended up with a hocon config.

kliushnichenko:
docs and some minor improvements added.  
The major change is the removal of a `type` property in custom problem errors. Since JSON Pointer spec clearly defines how to point to the whole document(a `""` pointer). Ready to merge

docs/asciidoc/problem-details.adoc
```diff
@@ -10,52 +10,44 @@ Thankfully, there’s a standard called https://www.rfc-editor.org/rfc/rfc7807[I
...
-{
-  ...  
-  error(new ProblemDetailsHandler()          // <1>
-    .log4xxErrors()                          // <2>
-    .mute(StatusCode.UNAUTHORIZED)           // <3>
-  )
-}
-----
+<1> By default, only server errors (5xx) will be logged. You can optionally enable the logging of client errors (4xx). If `DEBUG` logging level is enabled, the log will contain a stacktrace as well.
+<2> You can optionally mute some status codes completely.
```
jknack, line 43: Let's add a note about what `mute` means, like `debug` log statement.

jooby/src/main/java/io/jooby/problem/ProblemDetailsHandler.java
```diff
@@ -192,4 +201,25 @@ private void logProblem(Context ctx, HttpProblem problem, Throwable cause) {
   private String buildLogMsg(Context ctx, HttpProblem problem, StatusCode statusCode) {
     return "%s | %s".formatted(ErrorHandler.errorMessage(ctx, statusCode), problem.toString());
   }
+
+  public static ProblemDetailsHandler fromConfig(Config config) {
```
jknack, line 205: rename to `from` and let's add javadoc

jooby/src/main/java/io/jooby/validation/JsonPointer.java
```diff
@@ -0,0 +1,44 @@
...
+ * Transforms hibernate-validator (or avaje-validator) `propertyPath` into
+ * <a href="https://www.rfc-editor.org/rfc/rfc6901.html">JSON POINTER</a> format.
+ * For example:
+ * <p>"person.firstName" -> "/person/firstName"</p>
+ * <p>"persons[0].firstName" -> "/persons/0/firstName"</p>
+ *
+ * @author kliushnichenko
+ * @since 3.4.2
+ */
+public class JsonPointer {
```
jknack, line 18: cool

jknack (changes requested):
just few minor changes

docs/asciidoc/problem-details.adoc
kliushnichenko: The description is actually correct. `ProblemDetailsHandler` redefines the logging approach. Muted codes/types aren't logged under any condition, and here is why. From the app perspective the reason for `mute` can be some sort of security requirement. Let's take a look at the RFC7807 example   
image

I'm not sure if such sensitive user data like an account and its balance should be logged even under the `debug` level. That's why `mute` here means mute completely.

But we still have enough options to control logging:
- `log4xxErrors` flag to enable the logging of all client errors
- with `DEBUG` level the stack trace is also included when  some investigation is needed
- `mute` as a pinpoint mechanism to hide sensitive or annoying logs completely

So, do you want me to return the `debug` logging for `mute`?

jknack (approved):

PR #3563, opened 2024-10-19
annotation processors order note

kliushnichenko:
After `jooby-apt` rewrite the annotation-processors order becomes quite important. Added a note in the docs to save some brain cells for newbies.


jknack:
Thank you. Isn't any way to improve it?

jknack (approved):
Let's add the exact same comment as `NOTE:` in the adoc (in addition to what you did on xml)

kliushnichenko:
done, pls check

jknack (approved):

PR #3638, opened 2025-03-31
fix flyway baseline

gleb619:
With configuration like: 
```
flyway.run=[baseline, migrate]
```
application will crash with message '_Unknown flyway command: baseline_' but still provide baseline to db. So we need a fix)


gleb619:
@jknack please review, Is there any additional information needed?

jknack (approved):

Issue #3662, opened 2025-05-19
Reproducible Builds


froque:
The project at https://github.com/jvm-repo-rebuild/reproducible-central identified somes issues while trying to reproduce this project:

> koFiles="jooby-graphiql-3.8.1.jar jooby-cli-3.8.1.jar jooby-cli-3.8.1.zip"

https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/jooby/jooby-project-3.8.1.buildcompare

I investigated a bit and these are due to 2 issues:
 1. a lack of `package-lock.json` files (removed in 86dd4cfa7c5631973e7b5b98c5e905596d5f7252)
 1. writing `dependencies.properties` using full paths 

To solve 1. the `package-lock.json` files should be added back.

To solve 2. it might be possible to use https://github.com/mojohaus/properties-maven-plugin/pull/126 to exclude `argLine`. This is not yet published though. 

### Related issues
  - https://github.com/jooby-project/jooby/issues/2670
  - https://github.com/jooby-project/jooby/pull/3585

jknack:
can you explain the 2nd issue? I didn't get it.

froque:
Sure

It is the same issue as in this v3.5.4: https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/jooby/jooby-project-3.5.4.diffoscope

```
├── dependencies.properties
│ @@ -1,12 +1,12 @@
│  #Properties
│  HikariCP.version=6.2.0
│  Module-Name=io.jooby.cli
│  archetype-packaging.version=3.2.1
│ -argLine=-javaagent\:/home/runner/.m2/repository/org/jacoco/org.jacoco.agent/0.8.12/org.jacoco.agent-0.8.12-runtime.jar\=destfile\=/home/runner/work/jooby/jooby/modules/jooby-cli/target/jacoco.exec
│ +argLine=-javaagent\:/var/maven/.m2/repository-0002/org/jacoco/org.jacoco.agent/0.8.12/org.jacoco.agent-0.8.12-runtime.jar\=destfile\=/var/maven/app/modules/jooby-cli/target/jacoco.exec
│  asm.version=9.7.1
│  assertj-core.version=3.26.3
│  auto-service.version=1.1.1
```

The usage of properties-maven-plugin in https://github.com/jooby-project/jooby/blob/3.x/modules/jooby-cli/pom.xml#L99 writes `argLine` into `dependencies.properties` with full paths

[...]

PR #3681, opened 2025-05-27
drop unused property

hboutemy:
will solve the last Reproducible Builds issue after #3662 
see https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/jooby/README.md


jknack (approved):

hboutemy:
I confirm, everything is ok now https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/jooby/README.md
party time!

PR #3700, opened 2025-06-23
Fix Javadoc links on the site

lewisbirks:
I was having a look over the Jooby documentation on [jooby.io](https://jooby.io/) and noticed that when linking to methods in the javadoc the links weren't working as expected. For example in the [Context](https://jooby.io/#context) section there is a link to the `Context#locales()` method, the generated link for this is
`https://www.javadoc.io/doc/io.jooby/jooby/latest/io.jooby/io/jooby/Context.html#locales--`
when it needs to be
`https://www.javadoc.io/doc/io.jooby/jooby/latest/io.jooby/io/jooby/Context.html#locales()`

I have updated the JavadocProcessor to handle this as well as handling links to jooby javadoc from other artifacts (for example when linking to the netty server implementation in the [Worker Executor](https://jooby.io/#execution-model-worker-executor) section)

I've had to introduce another attribute, `module`, as well for the processor when the maven artifact doesn't line up with the declared module. This is only used in the [DB Scheduler](https://jooby.io/modules/db-scheduler/) section for now but if the documentation was updated to include javadoc links to artifacts such as `jooby-awssdk-v[1|2]` or `jooby-commons-email` then it would be used there as well.

Other minor fixes have also been applied where noticed.

### DIFF
--- a/docs/asciidoc/handlers/ssl.adoc
+++ b/docs/asciidoc/handlers/ssl.adoc
@@ -85,9 +85,8 @@ import io.jooby.handler.SSHandler
 
 For more information about SSL, please check the <<server-https-support, configure SSL>> section.
 
-// code review comment, this referenced doc doesn't exist
 [TIP]
 ====
 If you run behind a reverse proxy that has been configured to send the X-Forwarded-* header,
-please consider to add the link:proxy-peer-address.adoc[] to your pipeline.
+please consider to add the <<router-options-trust-proxy, trust proxy>> to your pipeline.
 ====
--- a/docs/asciidoc/handlers/ssl.adoc
+++ b/docs/asciidoc/handlers/ssl.adoc
@@ -1,12 +1,12 @@
 === SSLHandler
 
-The javadoc:SSLHandler[] forces client to use HTTPS by redirecting non-HTTPS calls to the HTTPS version.
+The javadoc:handler.SSLHandler[] forces client to use HTTPS by redirecting non-HTTPS calls to the HTTPS version.
 
 .Force SSL
 [source, java, role = "primary"]
 ----
 import io.jooby.Jooby;
-import io.jooby.SSLHandler;
+import io.jooby.handler.SSLHandler;
 ...
 {
   
@@ -24,7 +24,7 @@ import io.jooby.SSLHandler;
 [source, kotlin, role = "secondary"]
 ----
 import io.jooby.Jooby
-import io.jooby.SSHandler
+import io.jooby.handler.SSHandler
 ...
 {
   serverOptions {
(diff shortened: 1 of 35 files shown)


docs/asciidoc/handlers/ssl.adoc
```diff
@@ -85,6 +85,7 @@ import io.jooby.SSHandler
 
 For more information about SSL, please check the <<server-https-support, configure SSL>> section.
 
+// code review comment, this referenced doc doesn't exist
 [TIP]
 ====
 If you run behind a reverse proxy that has been configured to send the X-Forwarded-* header,
```
lewisbirks, line 91: There is a link to `link:proxy-peer-address.adoc[]` which doesn't exist, should the tip be removed or is there a more appropriate place to link to?

docs/asciidoc/modules/graphql.adoc
```diff
@@ -254,4 +254,5 @@ https://github.com/prisma-labs/graphql-playground[GraphQL Playground] should be
 
 === Starter
 
+// Code review question: should this remain 2.x?
 Checkout the starter/demo project for GraphQL: https://github.com/jooby-project/jooby/tree/2.x/starters/graphql-starter[GraphQL Starter].
```
lewisbirks, line 258: There are a few references to `2.x` starters, should these be updated to `3.x`?

docs/asciidoc/responses.adoc
```diff
@@ -773,10 +773,10 @@ Family of send methods include:
 
 - javadoc:Context[send, byte[\]]
 - javadoc:Route[consumes, io.jooby.MediaType...]
-- javadoc:Context[send, java.nio.Buffer]
+// Code review comment: is this supposed to be jooby databuffer or nio bytebuffer
+- javadoc:Context[send, java.nio.ByteBuffer]
```
lewisbirks, line 777: There are many overloads here, which version is wanted or do we want all of the overloads?

lewisbirks (commented):
Just some questions for when the code is being reviewed 🙂

docs/asciidoc/handlers/ssl.adoc
jknack: it must be: <<router-trust-proxy, trust proxy>>

jknack:
@lewisbirks Thank you! Appreciated the time and fixes.

docs/asciidoc/responses.adoc
lewisbirks: This is what it would look like with all the `send` overloads taken into account

```asciidoc
- javadoc:Context[send, byte[\]]
- javadoc:Context[send, byte[\]...]
- javadoc:Context[send, io.jooby.buffer.DataBuffer]
- javadoc:Context[send, io.jooby.FileDownload]
- javadoc:Context[send, io.jooby.StatusCode]
- javadoc:Context[send, java.io.InputStream]
- javadoc:Context[send, java.lang.String]
- javadoc:Context[send, java.lang.String, java.nio.charset.Charset]
- javadoc:Context[send, java.nio.ByteBuffer]
- javadoc:Context[send, java.nio.ByteBuffer[\]]
- javadoc:Context[send, java.nio.channels.FileChannel]
- javadoc:Context[send, java.nio.channels.ReadableByteChannel]
- javadoc:Context[send, java.nio.file.Path]
```
image

lewisbirks: added as part of a54ace5

PR #3709, opened 2025-07-01
Fix anchor links being overwritten

lewisbirks:
I noticed that the subsection [MVC API - Execution Model](https://jooby.io/#mvc-api-execution-model) was being used as the anchor link in the content for the [Execution Model](https://jooby.io/#execution-model) heading.

I think this is due to the `tocItems` method updating all links in the documents when the id changes rather than just those within it's element
https://github.com/jooby-project/jooby/blob/5e345b59719ad18317ada7e2a7928e164d611e07/docs/src/main/java/io/jooby/adoc/DocGenerator.java#L411-L420

I have updated this method to use `h` rather than `doc` and from what I can tell from a manual pass through, all links seem to work.

I've also moved the post processing functionality to make use of the inbuilt AsciiDoctorJ [post processing extension](https://docs.asciidoctor.org/asciidoctorj/latest/extensions/postprocessor/), this should make the generation process slightly more efficient as it cuts down on the amount of file reads/writes being performed. As this is a separate change that is only tangentially related by being in the same area I can move it to it's own PR if wanted.

### DIFF
--- a/docs/asciidoc/packaging/packaging.adoc
+++ b/docs/asciidoc/packaging/packaging.adoc
@@ -9,7 +9,7 @@ application.
 
 [TIP]
 ====
-The https://jooby.io/#getting-started[jooby-cli] takes care of configures everything for single jar
+The link:/#getting-started[jooby-cli] takes care of configures everything for single jar
 distribution. Next example shows how to do it in case you created your application manually.
 ====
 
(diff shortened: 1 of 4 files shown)


jknack (approved):
Thank you! Any enhancement here is always welcome.

Issue #3729, opened 2025-07-22
open-api: parse javadoc


jknack:
Find a way to parse javadoc as part of open-api generation

jknack:
oh man, that is awesome. I didn't know it.

agentgt:
Indeed it is awesome. I use it to [generate builders from a single static factory method](https://jstach.io/doc/rainbowgum/current/apidocs/io.jstach.rainbowgum.annotation/io/jstach/rainbowgum/annotation/LogConfigurable.html) and use the Javadoc as well as null annotations to figure things out [(example of generated file whose javadoc is based on the static method)](https://jstach.io/doc/rainbowgum/current/apidocs/io.jstach.rainbowgum/io/jstach/rainbowgum/output/FileOutputBuilder.html). I can't remember if I told @SentryMan or vice versa about but apparently their parser is for more [sophisticated than mine](https://github.com/jstachio/rainbowgum/blob/3c777286a1ebf3f3a6028b63bdd7dc27e661dfa5/rainbowgum-apt/src/main/java/io/jstach/rainbowgum/apt/ConfigProcessor.java#L268).

Just be aware that if you target Java 25 which I can't remember if 4 is going for 21 or 25 that Markdown is now Javadoc option so you will need to check for that:

https://download.java.net/java/early_access/loom/docs/api/java.compiler/javax/lang/model/util/Elements.html#getDocCommentKind(javax.lang.model.element.Element)

SentryMan:
If it were up to me I would target 25. But I see the appeal of casting a wider net. Since you're a web library I would go no lower than 21 because of virtual threads.

PR #3733, opened 2025-07-27
WIP: open-api: parse javadoc

jknack:
- Java APT can't be used to read doc bc we use byte code analysis
- Use checkstyle for getting doc
- This feature will be only for Java code base (not available on Kotlin)
- ref #3729


jknack:
@SentryMan @agentgt probably the most complete open-api spec generator avaialble for Java.

PR #3738, opened 2025-07-29
cleanup NettyServer log

kliushnichenko:
BTW, the log showed that the server was instantiated twice, if it matters.


jknack:
sorry. how did you bootstrap? probably is the service loader static call from server options

kliushnichenko:
yeah, the static one is an extra call

PR #3844, opened 2026-01-15
jakarta default value support

kliushnichenko:
Adding support for Jakarta `@DefaultValue` annotation in MVC and open-api


jknack:
forgot `Named` was from `inject`

PR #3854, opened 2026-02-22
FEATURE: Unified JSON Projection API

jknack:
Provide a consistent way of filtering JSON output. This API allows developers to define exactly which fields of a Java object should be serialized, supporting both simple flat structures and deep nested graphs.

The API is designed to be library-agnostic, allowing the same projection definition to work across different JSON providers without requiring changes to domain models or POJOs.

Flexible Syntax: Supports Dot-notation (address.city) and Avaje/LinkedIn-style grouping (address(city, zip)).

Type-Safety: Full support for Java method references (User::getName) to provide refactor-safe projections.

Hierarchical Validation: Projections are validated against the target class hierarchy at definition time, including support for unwrapping Collections and Maps.

Fluent & Declarative API: A Projected<T> wrapper for programmatic use and a `@Project` annotation for MVC controllers.

- Programmatic (Script API):

```java
get("/user/{id}", ctx -> {
  User user = service.find(ctx.path("id").value());
  return Projected.wrap(user)
    .include("id, name")
    .include(User::getAddress, addr -> addr.include("city"));
});
```

- Declarative (MVC API):

```java
@get
@project("id, name, address(city)")
public User getUser(String id) {
  return service.find(id);
}
```

The feature consists of three primary components that work together to define, wrap, and apply the filtering logic.

1. Projection<T>

The engine that parses syntax and validates paths against the class hierarchy.

```java
// Manual definition
Projection<User> p = Projection.of(User.class)
    .include("id, address(city)");
```

2. Projected<T>

```java
// Script API
get("/user", ctx -> {
  return Projected.wrap(user).include(User::getName);
});
```

3. `@Project`

A declarative annotation for MVC controllers.

```java
// MVC API
@get
@project("id, name")
public User getUser() { ... }
```

### DIFF
--- a/docs/asciidoc/context.adoc
+++ b/docs/asciidoc/context.adoc
@@ -406,7 +406,7 @@ curl -d "id=root&pass=pwd" -X POST http://localhost:8080/user
 <2> Form as `multi-value map`  => `{id=root, pass=[pwd]}`
 <3> Form variable `id`         => `root`
 <4> Form variable `pass`       => `pwd`
-<5> Form as `U3853` object      => `User(id=root, pass=pwd)`
+<5> Form as `User` object      => `User(id=root, pass=pwd)`
 
 ==== Multipart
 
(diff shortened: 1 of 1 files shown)


jknack:
@kliushnichenko here is some

jknack:
@SentryMan I'm adding your json-view to jooby.

PR #3917, opened 2026-04-20
declarative websockets

kliushnichenko:
So,

```java
@WebSocketRoute("/chat/{username}")
public class ChatWebsocket {

  @OnConnect
  public String onConnect(WebSocket ws, Context ctx) {
    return "welcome";
  }

  @OnMessage
  public Map<String, String> onMessage(WebSocket ws, Context ctx, WebSocketMessage message) {
    return Map.of("echo", message.value());
  }

  @OnClose
  public void onClose(WebSocket ws, Context ctx, WebSocketCloseStatus status) {}

  @OnError
  public void onError(WebSocket ws, Context ctx, Throwable cause) {}
}
```

Will produce

```java
@io.jooby.annotation.Generated(ChatWebsocket.class)
public class ChatWebsocketWs_ implements io.jooby.Extension {
    protected java.util.function.Function<io.jooby.Context, ChatWebsocket> factory;

    public ChatWebsocketWs_() {
      this(io.jooby.SneakyThrows.singleton(ChatWebsocket::new));
    }

    public ChatWebsocketWs_(ChatWebsocket instance) {
      setup(ctx -> instance);
    }

    public ChatWebsocketWs_(io.jooby.SneakyThrows.Supplier<ChatWebsocket> provider) {
      setup(ctx -> provider.get());
    }

    public ChatWebsocketWs_(io.jooby.SneakyThrows.Function<Class<ChatWebsocket>, ChatWebsocket> provider) {
      setup(ctx -> provider.apply(ChatWebsocket.class));
    }

    private void setup(java.util.function.Function<io.jooby.Context, ChatWebsocket> factory) {
      this.factory = factory;
    }

    public void install(io.jooby.Jooby app) throws Exception {
      app.ws("/chat/{username}", this::wsInit);
    }

    private void wsInit(io.jooby.Context ctx, io.jooby.WebSocketConfigurer configurer) {
      /** See {@link ChatWebsocket#onConnect(io.jooby.WebSocket, io.jooby.Context)} */
      configurer.onConnect(ws -> {
        var c = this.factory.apply(ctx);
        var __wsReturn = c.onConnect(ws, ctx);
        ws.send(__wsReturn);
      });

      /** See {@link ChatWebsocket#onMessage(io.jooby.WebSocket, io.jooby.Context, io.jooby.WebSocketMessage)} */
      configurer.onMessage((ws, message) -> {
        var c = this.factory.apply(ctx);
        var __wsReturn = c.onMessage(ws, ctx, message);
        ws.render(__wsReturn);
      });

      /** See {@link ChatWebsocket#onClose(io.jooby.WebSocket, io.jooby.Context, io.jooby.WebSocketCloseStatus)} */
      configurer.onClose((ws, status) -> {
        var c = this.factory.apply(ctx);
        c.onClose(ws, ctx, status);
      });

      /** See {@link ChatWebsocket#onError(io.jooby.WebSocket, io.jooby.Context, Throwable)} */
      configurer.onError((ws, cause) -> {
        var c = this.factory.apply(ctx);
        c.onError(ws, ctx, cause);
      });
    }
}
```

and can be registered over

```java
{
    ws(new ChatWebsocketWs_());
}
```

Annotation name is `@WebSocketRoute` to avoid the clash with `io.jooby.WebSocket` interface.  
If we want a nice annotation like `@WebSocket` -> need to rename `io.jooby.WebSocket` into `io.jooby.WebSocketConnection` or something.


jknack:
Will, look closer in a bit. For now why not `@Path` instead of `@WebSocketRoute`. Also, should we do the same for SSE?

kliushnichenko:
>why not @Path instead of @WebSocketRoute

`@WebSocketRoute` makes it vividly recognizable as a WS handler, easier to catch in apt.

`@Path` can be. If `@Path` + `@OnConnect/@OnMessage` is sufficient to catch in apt

jknack:
Think path fit better and we don't introduce a new annotation.

jknack:
The router generators: `RestRouter, JsonRcpRouter, etc..` have a selection method have a look at them to figure it out how to do the Path + WebSocket annotation: 

- https://github.com/jooby-project/jooby/blob/main/modules/jooby-apt/src/main/java/io/jooby/internal/apt/JsonRpcRouter.java#L27
- https://github.com/jooby-project/jooby/blob/main/modules/jooby-apt/src/main/java/io/jooby/internal/apt/McpRouter.java#L42
- https://github.com/jooby-project/jooby/blob/main/modules/jooby-apt/src/main/java/io/jooby/internal/apt/RestRouter.java#L22

modules/jooby-apt/src/main/java/io/jooby/internal/apt/ws/WsRouter.java
```diff
@@ -0,0 +1,238 @@
...
+          "io.jooby.annotation.ws.OnConnect", WsLifecycle.CONNECT,
+          "io.jooby.annotation.ws.OnMessage", WsLifecycle.MESSAGE,
+          "io.jooby.annotation.ws.OnClose", WsLifecycle.CLOSE,
+          "io.jooby.annotation.ws.OnError", WsLifecycle.ERROR);
+
+  public WsRouter(MvcContext context, TypeElement clazz) {
+    super(context, clazz);
+  }
+
+  public static WsRouter parse(MvcContext context, TypeElement controller) {
```
jknack, line 39: here is your parse/selection method. Here you need to filter by `Path` + [WebSocket] annotation. Is `Path` required??

kliushnichenko:
upd:
- removed `@WebSocketRoute` in favor of `@Path`
- `@Path` is optional
- `@OnConnect` or `@OnMessage` annotation is the minimal required condition to treat the class as a WS handler

jknack:
looks good. Should we add support for parameter binding? Beside the WebSocketMessage, like `onMessage(MyBean bean)`; `onMessage(String message);`, etc... WebSocketMessage it is a io.jooby.value.Value we probably can reuse binding from Rest/MVC generator.

kliushnichenko:
> Should we add support for parameter binding? 

Good point, just added it, a bit hacky, but it seems to work fine, and no changes to MvcParameter

jknack (approved):
love it, thank you

PR #3977, opened 2026-07-12
persistent openapi specification

kliushnichenko:
Sometimes it is useful to have your OpenAPI spec persisted in the repository (in addition to what we have in the jar). Especially nowadays, within an agentic engineering approach.

So, introducing an optional property, `copyOpenApiSpecTo` to the Maven plugin generating the spec.
Works only if the property is present; it copies the same output to the specified path.


jknack (changes requested):
look good, we need to add the new option to gradle openapi task too.

kliushnichenko:
@jknack gradle support is in. The module doesn't seem to have any tests; guess we can ignore the failing codecov check

PR #3984, opened 2026-07-29
build: make schema properties order stable

kliushnichenko:
Now, with `openapi.yaml` in the repo, I also noticed that the spec is not always reproducible. A typical issue involves a `boolean` field/getter that causes properties to "jump" in the schema upon rebuild. This is a try to fix it without forcing an alphabetical order everywhere (trying to avoid an avalanche of changes in the existing OpenAPI specs)


jknack (approved):
never saw it. What OS and Java SDK are you running?

kliushnichenko:
Windows 11, OpenJdk 21. It literally pops up only for a boolean field with a getter named `is*`. Everything else works perfectly

